PR #116, opened 2019-11-29
Refactored Server.js to Close MongoDB Connection

### LINKED ISSUE #115: Refactor server.js to become a class
`requiring` modules by default in JavaScript [includes them as singletons](https://medium.com/@iaincollins/how-not-to-create-a-singleton-in-node-js-bd7fde5361f5). When we do:
```js
require('./server')
```
an http server instance is created, listening for requests indefinitely. This is bad for test cases as they will hang, making CI builds get stuck.

#116 Temporarily solves the hanging issue by calling a function to close the connection in the file that is last in alphabetical order for tests.

### To resolve this issue:
1. Make a constructor of some sorts that will create the server instance when called
2. Change all imports to be wrapped with `new`, e.g.
```js
var app = new (require('./server'))
```
3. Make sure that the function to disconnect is called at the end of every test file.

evanugarte:
This is done to eliminate the possiblity of any hanging connections as a result of running test cases. Each Mocha test file will create its own instance of a server before running any tests, and will close all connections after finishing.

Test case output:
image

Resolves #115

### DIFF
--- a/server.js
+++ b/server.js
@@ -121,9 +121,9 @@ class Server {
     return this.server
   }
 
-  closeConnection () {
+  closeConnection (done) {
     this.server.close()
-    this.mongoose.connection.close()
+    this.mongoose.connection.close(done)
   }
 }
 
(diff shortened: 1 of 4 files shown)


evanugarte:
This new script already caught an issue with #108. Cool! 
image

PR #130, opened 2019-11-30
Thai/officer manager API and 3d printing API

ThaiQ:
2 Things are done in this pull request:

1. Fixed API query parameter for 3d-Printing from name, email, and color to only date. This is done because the date is unique for each request to the nano-second.

2. Created a mongoose model and an API routing for the officer managing tool (did not test)


api/models/OfficerManager.js
```diff
@@ -0,0 +1,52 @@
+const mongoose = require('mongoose')
+const Schema = mongoose.Schema
+
+const OfficersSchema = new Schema(
```
evanugarte: why not call it `OfficerSchema`? The naming convention seems to be singular
image

api/models/OfficerManager.js
```diff
@@ -0,0 +1,52 @@
...
+  {
+    name: {
+      type: String,
+      required: true
+    },
+    email: {
+      type: String,
+      required: true
+    },
+    facebook: {
```
evanugarte: how about linkedin? we can also make this an issue and do it later after v1.0.

api/models/PrintingForm3D.js
```diff
@@ -34,7 +34,7 @@ const PrintingForm3DSchema = new Schema(
 
     // unique id for query
     // id = email of the user who requested
-    id: {
+    email: {
```
evanugarte: you will need to refactor the `3DPrintingForm` tests with this change. for example see [here](https://github.com/SCE-Development/Core-v4/blob/9127fd37323555dffde5644f9f19934d2fdbc9ed/test/3DPrintingForm.js#L72)

api/routes/3DPrintingForm.js
```diff
@@ -56,7 +56,7 @@ router.post('/submit', (req, res) => {
 router.post('/GetForm', (req, res) => {
   // Query Criteria, query all if empty
   let obj = {}
-  if (typeof req.body.id !== 'undefined') obj = { id: req.body.id }
+  if (typeof req.body.email !== 'undefined') obj = { email: req.body.email }
```
evanugarte: make 
```js
obj = { email: req.body.email }
```
on its own indented line

src/Pages/Login/Login.js
```diff
@@ -146,7 +146,7 @@ export default class extends Component {
       <Layout>
         <form onSubmit={this.handleSubmit}>
           <div id='body'>
-            <img id='img' src='images/SCE-glow.png' />
+            <img id='img' src='images/SCE-glow.png' alt='SCE Logo' />
```
evanugarte: already fixed the `alt` issue in #117, remove line and rebase

PR #142, opened 2019-12-20
New 3D API and New officer manager API

ThaiQ:
2 Things are done in this pull request:

Fixed API query parameter for 3d-Printing from name, email, and color to only date. This is done because the date is unique for each request to the nano-second.

Created a mongoose model and an API routing for the officer managing tool (did not test)


evanugarte:
This is good, but let's make two pull requests with this. One for 3D, one for officer.

PR #146, opened 2019-12-20
Added API schema, model, and routing for Officer Manger

ThaiQ:
Adding API schema, model, and routing for Officer Manager. 
With test-cases

### DIFF
--- a/api/models/OfficerManager.js
+++ b/api/models/OfficerManager.js
@@ -1,0 +1,52 @@
+const mongoose = require('mongoose')
+const Schema = mongoose.Schema
+
+const OfficerSchema = new Schema(
+  {
+    name: {
+      type: String,
+      required: true
+    },
+    email: {
+      type: String,
+      required: true
+    },
+    facebook: {
+      type: String,
+      required: false
+    },
+    github: {
+      type: String,
+      required: false
+    },
+    instagram: {
+      type: String,
+      required: false
+    },
+    // The access level is defined as follows:
+    // -2: Ban
+    // -1: Pending
+    // 0: Member
+    // 1: Officer
+    // 2: Admin
+    level: {
+      type: Number,
+      require: true
+    },
+    team: {
+      type: String,
+      require: true
+    },
+    quote: {
+      type: String,
+      require: false
+    },
+    major: {
+      type: String,
+      require: true
+    }
+  },
+  { collection: 'Officers' }
+)
+
+module.exports = mongoose.model('Officers', OfficerSchema)
--- a/api/models/OfficerManager.js
+++ b/api/models/OfficerManager.js
@@ -12,16 +12,13 @@ const OfficerSchema = new Schema(
       required: true
     },
     facebook: {
-      type: String,
-      required: false
+      type: String
     },
     github: {
-      type: String,
-      required: false
+      type: String
     },
     instagram: {
-      type: String,
-      required: false
+      type: String
     },
     // The access level is defined as follows:
     // -2: Ban
@@ -38,8 +35,7 @@ const OfficerSchema = new Schema(
       require: true
     },
     quote: {
-      type: String,
-      require: false
+      type: String
     },
     major: {
       type: String,
--- a/src/Pages/OfficerDB/App.js
+++ b/src/Pages/OfficerDB/App.js
@@ -1,0 +1,85 @@
+import React from 'react'
+import './App.css'
+import 'bootstrap/dist/css/bootstrap.min.css'
+import { Table } from 'reactstrap'
+
+export default class App extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      email: '',
+      password: '',
+      message: '',
+      AnimationCSS1: {},
+      AnimationCSS2: {},
+      people: [
+        {
+          firstName: 'Surabhi',
+          lastName: 'Gupta',
+          linkedIn: '@surabhig',
+          Github: 'surabhig',
+          description: '....',
+          major: 'Software Engineering'
+        },
+        {
+          firstName: 'Zach',
+          lastName: 'Menes',
+          linkedIn: '@zachymemes',
+          Github: 'zachMenes',
+          description: '..ggg..',
+          major: 'Computer Engineering'
+        }
+      ]
+    }
+  }
+
+  // displaying each row of the map, used for mapping
+  row = (person, index) => {
+    return (
+      <tr key={index}>
+        <td>{parseInt(index) + 1}</td>
+        <td>{person.firstName}</td>
+        <td>{person.lastName}</td>
+        <td>{person.linkedIn}</td>
+        <td>{person.Github}</td>
+        <td>{person.description}</td>
+        <td>{person.major}</td>
+      </tr>
+    )
+  }
+
+  render () {
+    return (
+      <div className='App'>
+        <link
+          rel='stylesheet'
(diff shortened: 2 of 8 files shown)


api/models/PrintingForm3D.js
```diff
@@ -34,7 +34,7 @@ const PrintingForm3DSchema = new Schema(
 
     // unique id for query
     // id = email of the user who requested
-    id: {
+    email: {
```
evanugarte: please rebase your branch. these changes were already merged in

evanugarte: @ThaiQ this branch is behind. please rebase it
image

api/models/OfficerManager.js
```diff
@@ -0,0 +1,52 @@
...
+      type: String,
+      required: true
+    },
+    email: {
+      type: String,
+      required: true
+    },
+    facebook: {
+      type: String,
+      required: false
```
evanugarte: the `required` field is by default false, remove if it isn't true

test/officerManager.js
```diff
@@ -0,0 +1,336 @@
...
+let serverInstance = null
+let app = null
+const expect = chai.expect
+// tools for testing
+const tools = require('../util/testing-utils/tools.js')
+
+chai.should()
+chai.use(chaiHttp)
+
+function initializeServer () {
```
evanugarte: In a seperate PR, we should throw `initializeServer` and `terminateServer` this into tools. I will open an issue for it.

test/officerManager.js
```diff
@@ -0,0 +1,336 @@
...
+    tools.emptySchema(User)
+    done()
+  })
+  after(done => {
+    terminateServer(done)
+  })
+
+  let token = ''
+
+  describe('/POST submit', () => {
```
evanugarte: for every API that requires a token, please also test that by submitting a form without a token will return `UNAUTHORIZED`.

test/officerManager.js
```diff
@@ -0,0 +1,336 @@
...
+    })
+
+    it('Should not return statusCode 200 when the required fields are not set', done => {
+      const form = {}
+      chai
+        .request(app)
+        .post('/api/officerManager/submit')
+        .send(form)
+        .then(function (res) {
+          expect(res).to.not.have.status(200)
```
evanugarte: change the field to be `BAD_REQUEST`. MongoDB APIs can also return 500 in some situations and it is best to test the defined behavior.

test/officerManager.js
```diff
@@ -0,0 +1,336 @@
...
+
+          done()
+        })
+        .catch(err => {
+          throw err
+        })
+    })
+  })
+
+  describe('/POST GetForm', () => {
```
evanugarte: for every API that requires a token, please also test that by submitting a form without a token will return `UNAUTHORIZED`.

test/officerManager.js
```diff
@@ -0,0 +1,336 @@
...
+      chai
+        .request(app)
+        .post('/api/officerManager/GetForm')
+        .send(form)
+        .then(function (res) {
+          expect(res).to.have.status(200)
+          res.body.should.be.a('array')
+          // make sure all child has querries parameter
+          res.body.forEach(obj => {
+            if (obj.email !== 'test@test.com') {
```
evanugarte: replace the body of `forEach` with 
```js
expect(obj.email).to.be.eql('test@test.com')
```

test/officerManager.js
```diff
@@ -0,0 +1,336 @@
...
+          expect(res).to.have.status(404)
+
+          done()
+        })
+        .catch(err => {
+          throw err
+        })
+    })
+
+    it('Should return statusCode 200 and a message if a form was edited', done => {
```
evanugarte: After updating, please verify that the email with this value can now be found. This is done after editing and deleting in the Event API, found [here](https://github.com/SCE-Development/Core-v4/blob/7eb49330115efe0619c97377f7b69382cca2d41d/test/Event.js#L214)

test/officerManager.js
```diff
@@ -0,0 +1,336 @@
...
+          expect(res).to.have.status(404)
+
+          done()
+        })
+        .catch(err => {
+          throw err
+        })
+    })
+
+    it('Should return statusCode 200 and a message if a form was deleted', done => {
```
evanugarte: After deleting, please verify that the database should now be empty. This is done after deleting in the Event API, found [here](https://github.com/SCE-Development/Core-v4/blob/7eb49330115efe0619c97377f7b69382cca2d41d/test/Event.js#L284)

test/officerManager.js
evanugarte: See #150

ThaiQ: can you work on that. I tried to pull them over to util but having issues with the variables

test/officerManager.js
ThaiQ: no token require for submit

test/officerManager.js
ThaiQ: changed to 400

test/officerManager.js
ThaiQ: no token is needed for get request

test/officerManager.js
```diff
@@ -0,0 +1,455 @@
...
+          expect(res).to.have.status(401)
+
+          done()
+        })
+        .catch(err => {
+          throw err
+        })
+    })
+
+    it('Should return an object of all forms', done => {
```
evanugarte: i would also add to the test case that the database returns the object you previously added. see [here](https://github.com/SCE-Development/Core-v4/blob/7eb49330115efe0619c97377f7b69382cca2d41d/test/Event.js#L148)

src/Pages/OfficerDB/App.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
+import './App.css'
+import 'bootstrap/dist/css/bootstrap.min.css'
+import { Table } from 'reactstrap'
+
+export default class App extends React.Component {
```
evanugarte: Please also write test cases for this component. You can use [this example](https://github.com/SCE-Development/Core-v4/blob/7eb49330115efe0619c97377f7b69382cca2d41d/test/frontend/Slideshow.test.js#L13) for reference.

test/frontend/OfficerDB.test.js
```diff
@@ -0,0 +1,23 @@
...
+import Enzyme, { mount } from 'enzyme'
+import { expect } from 'chai'
+
+import OfficerManager from '../../src/Pages/OfficerDB/App.js'
+import Adapter from 'enzyme-adapter-react-16'
+import { Table } from 'reactstrap'
+
+Enzyme.configure({ adapter: new Adapter() })
+
+describe('Officer Manager: <Table />', () => {
```
evanugarte: please keep the same [naming convention](https://github.com/SCE-Development/Core-v4/blob/7eb49330115efe0619c97377f7b69382cca2d41d/test/frontend/Slideshow.test.js#L13) for React tests.
```suggestion
describe('<OfficerManager />', () => {
```

test/frontend/OfficerDB.test.js
```diff
@@ -0,0 +1,23 @@
...
+  it('Should render a <Table /> component with one child', () => {
+    const wrapper = mount(<OfficerManager />)
+    expect(wrapper.find(Table)).to.have.lengthOf(1)
+  })
+
+  it('Should render a <tbody /> component with 2 child', () => {
```
evanugarte:
```suggestion
  it('Should render a <tbody /> component with 2 children', () => {
```

PR #155, opened 2019-12-22
Now follows naming conventions

### LINKED ISSUE #154: Fix React Test Naming Convention
- When #146 was merged in, the name of the test is called `<Table />`, when the function we are in fact testing is `<OfficerManager />`

- The naming convention for React test cases should be the name of the component that we are testing. The output of TravisCI API tests becomes confusing as `<Table />` is very abstract.
image

- The file line that needs to be changed is [here](https://github.com/SCE-Development/Core-v4/blob/3e23f1ec7a37880da92e3222c2e494ec6a155b67/test/frontend/OfficerDB.test.js#L13)

seemavora:
Closes #154

### DIFF
--- a/test/frontend/OfficerDB.test.js
+++ b/test/frontend/OfficerDB.test.js
@@ -10,8 +10,8 @@ import { Table } from 'reactstrap'
 
 Enzyme.configure({ adapter: new Adapter() })
 
-describe('<Table />', () => {
-  it('Should render a <Table /> component with one child', () => {
+describe('<OfficerManager />', () => {
+  it('Should render a <OfficerManager /> component with one child', () => {
     const wrapper = mount(<OfficerManager />)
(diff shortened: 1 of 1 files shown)


test/frontend/OfficerDB.test.js
```diff
@@ -10,8 +10,8 @@ import { Table } from 'reactstrap'
 
 Enzyme.configure({ adapter: new Adapter() })
 
-describe('<Table />', () => {
-  it('Should render a <Table /> component with one child', () => {
+describe('<OfficerManager />', () => {
+  it('Should render a <OfficerManager /> component with one child', () => {
```
evanugarte: Keep this line as `Table`, since that is what we are testing.
```suggestion
  it('Should render a <Table /> component with one child', () => {
```

PR #156, opened 2019-12-23
Added Hooks to Carousel

### LINKED ISSUE #148: Refactor Carousel Component
- remove `id` field from all items [here](https://github.com/SCE-Development/Core-v4/blob/3e8fc8c60200ddcf9c3d20be0d9defff1949078f/src/Components/Slideshow/Slideshow.js#L13)
- remove `caption` field from all items [here](https://github.com/SCE-Development/Core-v4/blob/3e8fc8c60200ddcf9c3d20be0d9defff1949078f/src/Components/Slideshow/Slideshow.js#L17)
- use `index` instead of `item.id` [here](https://github.com/SCE-Development/Core-v4/blob/3e8fc8c60200ddcf9c3d20be0d9defff1949078f/src/Components/Slideshow/Slideshow.js#L141), since we removed it.
- make the component use hooks instead of classes [following this guide](https://alligator.io/react/converting-to-a-hook/)

### After our meeting today
- Remove bind for each function [like this](https://stackoverflow.com/a/32192892)

seemavora:
Resolves #148


src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
+          directionText='Next'
+          onClickHandler={handleNext}
+        />
+      </Carousel>
+    </div>
+  )
 }
 
-export default Slideshow
+export default Slideshow // exposes Slideshow to other modules
```
evanugarte:
```suggestion
export default Slideshow
```

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
-      )
-    })
-
-    return (
-      <div id='clicker'>
-        <style>
-          {`.slideshow {
+  return (
+    <div id='clicker'>
+      <style>
```
evanugarte: can you move the contents of this `<style>` tag to Slideshow.css?

```css
.slideshow {
 align-items: center;
 height: 94vh;
}
```

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
+function Slideshow () {
+  const [currentSlide, setCurrentSlide] = useState(0)
+  const [animating, setAnimating] = useState(false)
+  function handleOnExiting () {
+    setAnimating(true)
   }
 
-  handleOnExiting () {
-    this.animating = true
+  function handleOnExited () {
```
evanugarte: remove this function, see comment below for how to substitute it

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
-    this.handlePrevious = this.handlePrevious.bind(this)
-    this.onNext = this.handleNext.bind(this)
-    this.onPrevious = this.handlePrevious.bind(this)
-    this.handleGoToIndex = this.handleGoToIndex.bind(this)
-    this.handleOnExiting = this.handleOnExiting.bind(this)
-    this.handleOnExited = this.handleOnExited.bind(this)
+function Slideshow () {
+  const [currentSlide, setCurrentSlide] = useState(0)
+  const [animating, setAnimating] = useState(false)
+  function handleOnExiting () {
```
evanugarte: remove this function too. both `handleOnExiting` and `handleOnExited` do the same thing.

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
+          onClickHandler={handleGoToIndex}
+        />
+        {items.map((item, index) => {
+          return (
+            <CarouselItem
+              id={'caption-' + index}
+              className='slideshow text-center'
+              tag='div'
+              key={index}
+              onExiting={handleOnExiting}
```
evanugarte:
```suggestion
              onExiting={() => setAnimating(false)}
```

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
+              onExited={handleOnExited}
```
evanugarte:
```suggestion
              onExited={() => setAnimating(false)}
```

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
-            direction='next'
-            directionText='Next'
-            onClickHandler={this.handleNext}
-          />
-        </Carousel>
-      </div>
-    )
-  }
+      </style>
+      <Carousel activeIndex={currentSlide} next={onNext} previous={onPrevious}>
```
evanugarte:
```suggestion
      <Carousel activeIndex={currentSlide} next={handleNext} previous={handlePrevious}>
```

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
 
-  handleOnExiting () {
-    this.animating = true
+  function handleOnExited () {
+    setAnimating(false)
   }
 
-  handleOnExited () {
-    this.animating = false
+  function onNext () {
```
evanugarte: this function should be removed, all it does is forward the call to `handleNext()`

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,140 @@ import './Slideshow.css'
...
+    handleNext()
   }
 
-  onNext () {
-    this.handleNext()
+  function onPrevious () {
```
evanugarte: this function should be removed, all it does is forward the call to `handlePrevious()`

src/Components/Slideshow/Slideshow.js
seemavora: I believe this would be onExiting-{() => setAnimating(true)}  since previously handOnExiting was set to true. Is that correct?

src/Components/Slideshow/Slideshow.css
```diff
@@ -17,3 +17,8 @@
 #clicker .carousel-indicators li{
   background-color: black;
 }
+
+.slideshow{
+  align-items: center;
+  height: 94vh;
+}
```
evanugarte: add a blank newline at the bottom, [this post](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) explains why we do it

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,121 @@ import './Slideshow.css'
...
+    if (animating) return
+    const nextIndex = currentSlide === items.length - 1 ? 0 : currentSlide + 1
+    setCurrentSlide(nextIndex)
   }
 
-  handleOnExiting () {
-    this.animating = true
+  function handlePrevious () {
+    if (animating) return
+    const nextIndex = currentSlide === 0 ? items.length - 1 : currentSlide - 1
```
evanugarte: we can use modulo to "wrap around" the index

Take a look at what `Line 75-76` is doing, and how we can use mod to refactor this

hint:
image

src/Components/Slideshow/Slideshow.js
```diff
@@ -10,185 +10,121 @@ import './Slideshow.css'
...
-    this.onPrevious = this.handlePrevious.bind(this)
-    this.handleGoToIndex = this.handleGoToIndex.bind(this)
-    this.handleOnExiting = this.handleOnExiting.bind(this)
-    this.handleOnExited = this.handleOnExited.bind(this)
+function Slideshow () {
+  const [currentSlide, setCurrentSlide] = useState(0)
+  const [animating, setAnimating] = useState(false)
+  function handleNext () {
+    if (animating) return
+    const nextIndex = currentSlide === items.length - 1 ? 0 : currentSlide + 1
```
evanugarte: As mentioned below, we can use modulo here as well

PR #159, opened 2019-12-23
Abstracted Server Initialization and Termination

### LINKED ISSUE #150: Abstract Server Initialization and Termination for Tests
The functions `initializeServer()` and `terminateServer()` pop up in [each test case](https://github.com/SCE-Development/Core-v4/blob/7eb49330115efe0619c97377f7b69382cca2d41d/test/Event.js#L19). The two functions should now be moved to [util/testing-utils/tools.js](https://github.com/SCE-Development/Core-v4/blob/7eb49330115efe0619c97377f7b69382cca2d41d/util/testing-utils/tools.js) to abstract their behavior.

nanarboursal:
i moved the function woohoo winter break

Closes #150

### DIFF
--- a/util/testing-utils/tools.js
+++ b/util/testing-utils/tools.js
@@ -6,7 +6,22 @@ function emptySchema (schema) {
   })
 }
 
+const server = require('../../server')
+let serverInstance = null
+
+function initializeServer () {
+  serverInstance = new server.Server()
+  serverInstance.openConnection()
+  return serverInstance.getServerInstance()
+}
+
+function terminateServer (done) {
+  serverInstance.closeConnection(done)
+}
+
 // Exporting functions
 module.exports = {
-  emptySchema: emptySchema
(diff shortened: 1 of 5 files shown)


util/testing-utils/tools.js
```diff
@@ -6,7 +6,25 @@ function emptySchema (schema) {
...
 }
 
+const server = require('../../server')
+let serverInstance = null
+let app = null
+
+function initializeServer () {
+  serverInstance = new server.Server()
+  serverInstance.openConnection()
+  app = serverInstance.getServerInstance()
```
evanugarte:
```suggestion
  return serverInstance.getServerInstance()
```

util/testing-utils/tools.js
```diff
@@ -6,7 +6,25 @@ function emptySchema (schema) {
...
+  return app
```
evanugarte:
```suggestion
```

PR #171, opened 2019-12-26
 Refactored Routing + Navbar Rendering

evanugarte:
We now check for the user's auth status in the highest component and use a
custom PrivateRoute component for securing routes. We also now use array.map in a seperate file, Routing.js to render the routes

Each component in Routing.js is wrapped by the NavbarWrapper
component. NavbarWrapper renders the correct component based on the
route and user's account priviledge level.

we now have collapsable navbars!
image


ThaiQ:
'Snack and Food' is repeated, the 2nd 'Snack and Food' should be The Team since it routes to '/Team'
image

Dashboard is messed up, cannot click on its nav bar
image

Dont need this since 'Home' go back home
image

ThaiQ:
Please keep it blue, grey just not fit with a lot of our design
image

Does the logout button suppose to be like that?
image

Drop downs are hidden
image

ThaiQ:
they are all shifted left now. They should shift right as the home page nav
image

evanugarte:
@ThaiQ fixed. please also add review comments from now on with the green button on the right side:
image

ThaiQ:
I know i just dont :))

PR #180, opened 2019-12-29
Member Profile

ThaiQ:
Membership Profile Page


src/Pages/Profile/MemberView/InfoCard.js
```diff
@@ -0,0 +1,35 @@
...
+import { Input } from 'reactstrap'
+import './ProfileModifier.css'
+
+export default function ProfileCard (props) {
+  const [month] = useState(parseInt(new Date().getMonth()) + 1)
+  const [pic] = useState(getPic())
+
+  function getPic () {
+    let pic
+    if (month === 3 || month === 4 || month === 5) { pic = require('./Image/spring.jpg') } else if (month === 6 || month === 7 || month === 8) { pic = require('./Image/summer.jpeg') } else if (month === 9 || month === 10 || month === 11) { pic = require('./Image/fall.jpeg') } else if (month === 12 || month === 1 || month === 2) { pic = require('./Image/winter.jpeg') }
```
evanugarte: None of this!
Instead, make another `.js` file in the `MemberView/` directory that will have a function like below:
```js
export default function getPicBySeason() {
  // code here
}
```

Then, just import it in this file and call on it. 

A suggestion for this exported function's body:
```js
import fall from './Image/fall.jpeg'
// import the rest of the seasons like above


export default function getPicBySeason() {
  let pic = null
  if (month === 9 || month === 10 || month === 11) {
    pic = fall 
  } // else if (...)
  
  return pic
}

File: src/Pages/Profile/MemberView/InfoCard.js [hunk 2]
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react'
+import { Input } from 'reactstrap'
+import './ProfileModifier.css'
+
+export default function ProfileCard (props) {
+  const [month] = useState(parseInt(new Date().getMonth()) + 1)
```
evanugarte: You don't need a hook here, and with my suggestion for the date function you can remove this entire line.

src/Pages/Profile/MemberView/InfoCard.js
```diff
@@ -0,0 +1,35 @@
...
+  const [pic] = useState(getPic())
```
evanugarte: With my suggestion for the date function you can remove this entire line.

src/Pages/Profile/MemberView/InfoCard.js
```diff
@@ -0,0 +1,35 @@
...
+      <img id='clip' alt='side' src={pic} />
+      <img id='clip2' alt='side' src={pic} />
+      <div id='profileBox'>
+        {props.fields.map((field, ind) => (
+          <h3 key={ind} id='innerTextTop'>
+            <b>{field.front}:</b> {field.end}
+          </h3>
+        ))}
+        <h3 id='innerTextTop'>
+          New Password: <Input type='password' />
```
evanugarte: While we don't need a complete component, we should at least store the password in a hook.

up top:
```js
const [password, setPassword] = useState('')
```

then add an `onChange` prop to your `Input` to set the password.

src/Pages/Profile/MemberView/InfoCard.js
```diff
@@ -0,0 +1,35 @@
...
+        </h3>
+        <h3 id='innerTextTop'>
+          Confirm Password: <Input type='password' />
```
evanugarte: Store the confirmed password in a hook.

up top:
```js
const [confirmedPassword, setConfirmedPassword] = useState('')
```

then add an `onChange` prop to your `Input` to set the hook.

src/Pages/Profile/MemberView/InfoCard.js
```diff
@@ -0,0 +1,35 @@
...
+          <button style={{ marginTop: '10px' }}>Change Password</button>
```
evanugarte: This button should be disabled until the `password` and `confirmedPassword` hooks are equal.

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from 'react';
+import './ProfileModifier.css';
+import Layout from '../../../Components/Layout/Layout'
+import Footer from '../../../Components/Footer/Footer.js'
+import InfoCard from './InfoCard.js'
+
+export default class Profile extends Component {
+
+  constructor(props) {
```
evanugarte: remove the constructor as there is nothing in it.

src/Pages/Profile/MemberView/InfoCard.js
evanugarte: By the way, `getMonth` is [indexed at zero](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth), keep that in mind

ThaiQ:
This is in progress. That's just the UI. I need the other 3 PR to merge before I can continue

src/Pages/Profile/MemberView/getPicBySeason.js
```diff
@@ -0,0 +1,19 @@
+import spring from './Image/spring.jpg'
+import fall from './Image/fall.jpeg'
+import winter from './Image/winter.jpeg'
+import summer from './Image/summer.jpeg'
+
+export function getPicture () {
```
evanugarte:
```suggestion
export function getPictureByMonth () {
```

src/Pages/Profile/MemberView/ProfileModifier.css
```diff
@@ -0,0 +1,37 @@
...
+}
+
+#clip2 {
+  position: relative;
+  width: 40%;
+  height: 40%;
+  right: 5%;
+  display: inline-block;
+  clip-path: polygon(100% 0, 100% 100%, 0 100%);
+}
```
evanugarte: keep blank newline at bottom of the file

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -0,0 +1,46 @@
...
+          }}
+          src='images/SCE-glow.png' />
+
+        <InfoCard fields={fields} />
+
+        <Footer />
+      </Layout>
+    );
+  }
+}
```
evanugarte: keep blank newline

src/Pages/Profile/MemberView/ProfileModifier.css
```diff
@@ -0,0 +1,37 @@
+#enclose {
```
evanugarte: Rename this file to use lower case and dashes. An explanation can be found [here](https://www.lifewire.com/naming-css-style-sheet-files-3466881)

src/Routing.js
```diff
@@ -90,6 +91,12 @@ export default function Routing ({ appProps }) {
       path: '/register',
       allowedIf: !userIsAuthenticated,
       redirect: '/'
+    },
+    {
+      Component: Profile,
```
evanugarte: Good to see that we are routing more scalable now, nice!

src/Routing.js
```diff
@@ -121,7 +128,8 @@ export default function Routing ({ appProps }) {
                     enableAdminNavbar={
                       userIsOfficer &&
                       path !== '/2DPrinting' &&
-                      path !== '/3DPrintingForm'
+                      path !== '/3DPrintingForm' &&
```
evanugarte: befor the `enableAdminNavbar` prop becomes super long with special routes, lets add an optional field to `signedInRoutes`

```js
// ...
{
  Component: Login,
  path: '/login',
  allowedIf: !userIsAuthenticated,
  redirect: '/',
  // add the prop below
  disableAdminNavbar: true
}
```

then, the prop for `NavbarWrapper` can become
```jsx
enableAdminNavbar={!disableAdminNavbar}
```

evanugarte: You would also need to change the `array.map` parameters above, adding `disableAdminNavbar` to it

test/frontend/UserProfile.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+  it('Should render a <img /> where src = public/images/SCE-glow.png', () => {
+    expect(wrapper.find('img').get(0).props.src).equals('images/SCE-glow.png')
+  })
+
+  it('Should render 2 <img /> where src = spring, winter, fall, or summer; bases from month', () => {
+    let date
+    const RealDate = Date
+
+    function mockDate (isoDate) {
```
evanugarte: A better way to mock this would be to use sinon. We already have it as a dependency and an example of me using it is in #190 

e.g.
```js
describe('<UserProfile />', () => {
  var stub = null
  before(done => {
    // mock the getMonth date to return values we want
    stub = sinon.stub(Date, 'getMonth')
    stub.onCall(0).returns(1) // return the month of january
    stub.onCall(1).returns(3) // return the month of april
    stub.onCall(2).returns(6) // return august
    stub.onCall(3).returns(8) // return september
    done()
  })
  after(done => {
    // get rid of the stub
    if (stub) stub.restore()
    done()
  })
  // rest of your tests
})
```
Now, the first time you mount `UserProfile`, it should render winter from your mocked date. The next time should be spring, the next time... you get the idea.

Read more about `sub.onCall()` [here](https://sinonjs.org/releases/latest/stubs/#stuboncalln-added-in-v18)

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -0,0 +1,86 @@
...
+      .catch(err => {
+        console.log(err)
+      })
+  }
+
+  render() {
+    const fields =
+      (this.state.user) ?
+        [
+          { front: 'Name', end: this.state.fullName },
```
evanugarte: Let's rename the fields here to something more descriptive, e.g.
```suggestion
          { title: 'Name', value: this.state.fullName },
```

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -0,0 +1,86 @@
...
+          { front: 'Door Code', end: this.state.user.doorCode },
+          { front: 'Joined Date', end: this.state.user.joinDate.slice(0, 10) },
```
evanugarte: Is there a way we can get the join date without calling slice? Open an issue so that the API can return a formatted value.

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -0,0 +1,86 @@
...
+          { front: 'Email', end: this.state.user.email },
+          { front: 'Membership Expiration', end: this.state.user.membershipValidUntil.slice(0, 10) },
```
evanugarte: In the same issue previously described, add to it saying that we should return join and expiration dates that don't need to be sliced in the frontend.

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -0,0 +1,86 @@
...
+          { front: 'Email', end: '' },
+          { front: 'Membership Expiration', end: '' },
+        ]
+
+    return (
+      <div id='app'>
+        <h1 id='title'><b> SCE Member Profile </b></h1>
+
+        <img id='img' alt='sce logo'
+          style={{
```
evanugarte: This style should be inside a .css file

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -0,0 +1,86 @@
...
+          { front: '.', end: '' },
+          { front: 'Door Code', end: '' },
+          { front: 'Joined Date', end: '' },
+          { front: 'Email', end: '' },
+          { front: 'Membership Expiration', end: '' },
+        ]
+
+    return (
+      <div id='app'>
+        <h1 id='title'><b> SCE Member Profile </b></h1>
```
evanugarte: Do you need a bold tag? You should be able to do this with

```css
#id {
  font-weight: bold;
}
```

src/Routing.js
ThaiQ: this should be a different PR

ThaiQ: like navbar-refactor?

src/Pages/Profile/MemberView/Profile.jsx
ThaiQ: opened an issue

src/Pages/Profile/MemberView/InfoCard.js
```diff
@@ -0,0 +1,97 @@
...
+import { Input, Button } from 'reactstrap'
+import './profile_modifier.css'
+import axios from 'axios'
+
+const pic = require('./getPicBySeason')
+const bcrypt = require('bcrypt-nodejs')
+
+export default function ProfileCard (props) {
+  const [newPass, setNewPass] = useState(1)
+  const [confirmPass, setConfirmPass] = useState(2)
```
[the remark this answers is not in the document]
ThaiQ: yes, I just want to make sure they are different initially. will change it into something that make more sense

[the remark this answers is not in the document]
ThaiQ: Made it

ThaiQ:
Removed console.log(err). Nice mocking.

src/Pages/Profile/MemberView/profile_modifier.css
```diff
@@ -0,0 +1,49 @@
+#enclose {
```
evanugarte: remember the css naming convention? change this file name to use dashes.

PR #182, opened 2019-12-31
Officer Manager Board

ThaiQ:


src/Enums.js
```diff
@@ -6,6 +6,23 @@ export const membershipStatus = [
   { key: 'Admin', value: 2 }
 ]
 
+export const officerRole = [
```
evanugarte: To truly make this an enum, don't use an array. [This example](https://www.sohamkamani.com/blog/2017/08/21/enums-in-javascript/) helps

src/Enums.js
```diff
@@ -6,6 +6,23 @@ export const membershipStatus = [
...
+  'Development',
+  'Public Relation',
+  'Event Planning',
+  '3D-Printing'
+]
+
+export const chairRole = [
```
evanugarte: same here, see previous comment

src/Pages/OfficerDB/Edits/Adder.js
```diff
@@ -0,0 +1,215 @@
...
+        props.users.filter(user =>
+          user.firstName.toLowerCase().includes(query)
+        )
+      )
+      : setQueryUsers('')
+  }
+
+  async function submitOfficer () {
+    let accessLevel = ''
+    if (enums.getAllKeys(enums.officerRole).includes(role)) {
```
evanugarte: The additional functions in the enums file is getting confusing. By using numbers or constant strings, we don't have to worry about these things.

Sure we would store the role as 1, 2, or 3 but in the code they are reflected as constant values, e.g. DEVELOPMENT, EVENT_PLANNING, etc.

Can we change the enums to reflect this? See [here](https://github.com/SCE-Development/Core-v4/blob/b4707f9f5ef75f3806b7cc6d2cce5f81454e90f6/src/Enums.js#L6)

src/Pages/OfficerDB/Edits/ConfirmationModule.js
```diff
@@ -0,0 +1,55 @@
...
+import { ModalFooter, Button, Modal } from 'reactstrap'
+
+export default function Confirmation (props) {
+  const [toggleSubmit, setToggleSubmit] = useState(false)
+
+  return (
+    <ModalFooter>
+      <Button
+        onClick={async () => {
+          ;(await props.check()) && setToggleSubmit(!toggleSubmit)
```
evanugarte:
```suggestion
          (await props.check()) && setToggleSubmit(!toggleSubmit)
```

src/Pages/OfficerDB/Edits/Editor.js
```diff
@@ -0,0 +1,157 @@
...
+      placeholder: 'New github: ' + props.officer.github,
+      onChange: event => setGithub(event.target.value)
+    },
+    {
+      placeholder: 'New quote: ' + props.officer.quote,
+      onChange: event => setQuote(event.target.value)
+    }
+  ]
+
+  async function submitOfficer () {
```
evanugarte: this function should go inside `APIFunctions/`

src/Pages/OfficerDB/Edits/Adder.js
```diff
@@ -0,0 +1,215 @@
...
+    query.toLowerCase().trim()
+      ? setQueryUsers(
+        props.users.filter(user =>
+          user.firstName.toLowerCase().includes(query)
+        )
+      )
+      : setQueryUsers('')
+  }
+
+  async function submitOfficer () {
```
evanugarte: this function should go inside `APIFunctions/`

src/Pages/OfficerDB/InfoRow.js
```diff
@@ -0,0 +1,82 @@
+import React, { useState } from 'react'
+import 'bootstrap/dist/css/bootstrap.min.css'
+import { Modal, Button, ModalHeader, ModalBody } from 'reactstrap'
+import Editor from './Edits/Editor'
+
+const api = require('../../APIFunctions/OfficerDB')
```
evanugarte: lets not mix require with imports. instead, you can directly import the functions [like this](https://github.com/SCE-Development/Core-v4/blob/b4707f9f5ef75f3806b7cc6d2cce5f81454e90f6/src/Pages/Events/EventInfoModal.js#L12)

src/Pages/OfficerDB/InfoRow.js
```diff
@@ -0,0 +1,82 @@
...
+const svg = require('./SVG')
+
+export default function Row (props) {
```
evanugarte: Row is already a reactstrap component
```suggestion
export default function InfoRow (props) {
```

src/Pages/OfficerDB/OfficerDB.js
```diff
@@ -1,75 +1,111 @@
 import React from 'react'
-import './officer-db.css'
+import './OfficerDB.css'
 import 'bootstrap/dist/css/bootstrap.min.css'
-import { Table } from 'reactstrap'
+import { Table, Input } from 'reactstrap'
+import TableRow from './InfoRow'
```
evanugarte: Either rename the file as TableRow or do:
```suggestion
import InfoRow from './InfoRow'
```

src/Pages/OfficerDB/OfficerDB.js
```diff
@@ -1,75 +1,111 @@
...
             </tr>
           </thead>
+
           <tbody>
-            {this.state.people.map((person, index) => {
-              return this.row(person, index)
-            })}
+            {this.state.queryOfficers.map((officer, index) => (
+              <TableRow
+                officer={officer}
```
evanugarte: I am noticing that inside `TableRow` you are doing a lot of `props.officer.name`, `props.officer.role` etc.

To send the props over without being inside a `officer` object you can do

```suggestion
                {...officer}
```
you should then be able to access everything like `props.title`, `props.role`

src/Pages/OfficerDB/OfficerDB.js
```diff
@@ -1,75 +1,111 @@
...
+                user={this.props.user}
```
evanugarte: just send in the props like the previous comment
```suggestion
                {...this.props}
```

src/APIFunctions/Firebase.js
```diff
@@ -0,0 +1,18 @@
+import firebase from 'firebase/app'
```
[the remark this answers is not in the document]
ThaiQ: instead of ignore this file, lets keep it but commit an empty firebaseConfig

src/Pages/OfficerDB/Edits/Adder.js
ThaiQ: actually, the values indicate the access-level of each role.
If i use an object then I wouldn't be able to do ```js map ``` but I need it for the drop-downs

src/Pages/OfficerDB/Edits/Editor.js
ThaiQ: actually, that's my bad I renamed it into ```handleSubmitOfficer``` its not an axios function so i wouldn't put it in APIfunctions

src/Pages/OfficerDB/Edits/Adder.js
ThaiQ: Above comment

src/Pages/OfficerDB/OfficerDB.js
ThaiQ: Because i will need to pass on the entire ```officer``` object to another component, so it is simpler this way to use it directly from ```props.officer``` rather than smaller ```{...officer}```

src/Pages/OfficerDB/OfficerDB.js
ThaiQ: above comment

src/APIFunctions/Firebase.js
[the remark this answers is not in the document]
evanugarte: yes so remove these keys as anyone can see them now.

src/Pages/OfficerDB/Edits/Adder.js
evanugarte: Why not make one `Role` array, ranging from banned to executive?

ThaiQ: We could. I wanted to keep them separated from one another in case we need to use the chair array or the officer array separately in another component; so you dont get a bunch of irrelevant role when you only wanted chairs or officers

src/Enums.js
ThaiQ: then I can't map out the children

ThaiQ: My intention is that it allows me to map-out each child and can get each child's value

ThaiQ:
@evanugarte 

Is this where i add in the .gitignore file to ignore the file?
image

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
   { key: 'Admin', value: 2 }
 ]
 
+export const officerRole = {
```
evanugarte: Separate all the object fields with underscores for spaces.

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
...
+  DEVELOPMENT: 'Development',
+  PUBLICRELATION: 'Public Relation',
```
evanugarte: it is plural.
```suggestion
  PUBLIC_RELATIONS: 'Public Relations',
```

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
...
+  EVENTPLANNING: 'Event Planning',
```
evanugarte:
```suggestion
  EVENT_PLANNING: 'Event Planning',
```

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
...
+  PRINTING: '3D-Printing'
+}
+
+export const chairRole = {
+  PRESIDENT: 'President',
+  VICEPRESIDENT: 'Vice President',
```
evanugarte:
```suggestion
  VICE_PRESIDENT: 'Vice President',
```

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
...
+  TREASURER: 'Treasurer',
+  DEVELOPMENTCHAIR: 'Software Development Chair',
```
evanugarte:
```suggestion
  DEVELOPMENT_CHAIR: 'Software Development Chair',
```

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
...
+  EVENTPLANNINGCHAIR: 'Event Planning Chair',
```
evanugarte:
```suggestion
  EVENT_PLANNING_CHAIR: 'Event Planning Chair',
```

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
...
+  STOREMANAGER: 'Store Manager',
+  PRINTINGCHAIR: '3D-Printing/Hacking Station Chair'
```
evanugarte:
```suggestion
  PRINTING_SPECIALIST: '3D-Printing/Hacking Station Chair'
```
name it like this:
image

src/Enums.js
```diff
@@ -32,12 +32,31 @@ export const membershipStatus = [
...
+}
+
 export function getKey (enumArray, value) {
```
evanugarte: do you need getKey and getValue anymore?

evanugarte:
@ThaiQ Also squash your commits before re-requesting review

src/Enums.js
ThaiQ: Overview is using it, so yes

src/Pages/OfficerDB/Edits/Editor.js
```diff
@@ -0,0 +1,160 @@
...
+      pictureName: newPic || picture,
+      name: props.officer.name,
+      email: props.officer.email,
+      accessLevel: accessLevel,
+      linkedIn: linkedIn,
+      github: github,
+      quote: quote,
+      role: role
+    }
+    // Delete empty fields
```
evanugarte: why are you deleting it? you should just handle empty fields in your API [like this](https://github.com/SCE-Development/Core-v4/blob/89fbac4635e7ae72a54cc8d2a2cced59f071e708/api/routes/Event.js#L69)

src/Pages/OfficerDB/Edits/Uploader.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+const api = require('../../../APIFunctions/OfficerDB')
+
+let file = ''
+
+export default function Uploader () {
```
evanugarte: Id rather we not mix React components and utility functions in the same file, can you split this up? You can probably put the `Uploader` component back where it was.

src/Pages/OfficerDB/OfficerDB.js
evanugarte: hey man do this^^^^^^^^^^^^

PR #190, opened 2020-01-03
Added Status Code Constants & Token Functions

evanugarte:
This will reduce the number of repeated lines in tests and api/routes. The Token functions are
documented, tested and ensure that they work properly.

@ bad code
image

PR #191, opened 2020-01-03
Abstracted Server Responses & Token Verification

### LINKED ISSUE #192: Abstract HTTP Status Codes
[HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) in Core-v4 are often redefined (e.g. [here](https://github.com/SCE-Development/Core-v4/blob/a73e655fd9e9b4f7484f9abc0b1916548fd6d96f/api/routes/event.js#L13) and [here](https://github.com/SCE-Development/Core-v4/blob/a73e655fd9e9b4f7484f9abc0b1916548fd6d96f/api/routes/officerManager.js#L15)) or hardcoded (e.g. [here](https://github.com/SCE-Development/Core-v4/blob/a73e655fd9e9b4f7484f9abc0b1916548fd6d96f/test/Event.js#L46) and [here too](https://github.com/SCE-Development/Core-v4/blob/a73e655fd9e9b4f7484f9abc0b1916548fd6d96f/test/user.js#L104). We should abstract these codes, and instead import them for scalable use.

```js
const STATUS_CODES = {
  OK: 200,
  BAD_REQUEST: 400,
  UNAUTHORIZED: 401,
  FORBIDDEN: 403,
  NOT_FOUND: 404,
  CONFLICT: 409
}
```

evanugarte:
Using the work from #190, we checking if a token exists and if it is
valid through functions. We also no longer return hardcoded responses from APIs and check for them
in routes.
Resolves #192 #193

### DIFF
--- a/api/routes/3DPrintingForm.js
+++ b/api/routes/3DPrintingForm.js
@@ -1,35 +1,21 @@
-// PROJECT:   Core-v4
-//  Name:    R. Javier
-//  File:    api/routes/membershipApplication/index.js
-//  Date Created:  March 17, 2019
-//  Last Modified:  March 17, 2019
-//  Details:
-//      This file contains routing logic to service all routes requested under the the
-//                  "/memberApplication" endpoint (a.k.a. the Membership Application Module)
-//     which is publicly exposed to allow applications from the public facing site.
-//  Dependencies:
-//      JavaScript ECMAscript 6
-
 'use strict'
 
-// Includes (include as many as you need; the bare essentials are included here)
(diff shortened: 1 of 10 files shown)


ThaiQ:
Logout is no longer logout
image

Also, because of the API changes, to go src/Profile/AdminView.js/
change 
```js
  const [email, setEmail] = useState("")
to
  const [email, setEmail] = useState(props.user.email)
```

ThaiQ:
still can't logout from nav bar.

PR #197, opened 2020-01-04
Refactored Membership Page

### LINKED ISSUE #172: Fix Subscription Type Width
In the [MembershipPlan](https://github.com/SCE-Development/Core-v4/blob/dev/src/Pages/MembershipPlan/App.js) component, the width overflows, and the user can "scroll off" the webpage like below:

`localhost:3000/register`
image

We need to change the css on [this line](https://github.com/SCE-Development/Core-v4/blob/de206a9298f1968fd0c3fd5cc07c4a0cefb1e0bf/src/Pages/MembershipPlan/App.css#L3) for the width to remain the same. but not overflow off the regular page.

### LINKED ISSUE #196: Fix Navbar Overflow
Opening Chrome developer tools makes our navbar appear like so:
image

This is due to the `flex-flow` css attribute of reactstrap's `<Nav />` component. We can fix this issue with a custom css class.
image

evanugarte:
### See #208 first

1. Membership plan is now split into three views:
#### First View: Choose membership
image

#### Second View: Enter account info
image
image

#### Third View: Confirmation Page
image

### Selecing a Major
image
image
image

Resolves #172


ThaiQ:
images are stretched out make them thinner a little bit, add a little place between them please.
image

Can we have the fields (name, email, ...) float left? It looks better that way
image

evanugarte:
Gonna rewrite this component. @ whoever wrote it:
image

ThaiQ:
Edit button in profile/adminView is messed up
image

Why is that so fat sand stretchy?
image

Include middle initial because people might have the same first name or last name

evanugarte:
From now on, don't comment reviews anymore. Use the button so github will show that you review pull requests.

evanugarte:
@ThaiQ I think we should remove middle initial. Users can still have the same first last and middle names. Plus, we already uniquely identify them by email.

Facebook doesn't use middle initials:
image

Neither does Google:
image

Or Amazon:
image

ThaiQ:
my worry is when you have to search them up in a dashboard and 2-3 people pops up and the officer doesn't know who is who.

evanugarte:
We should differentiate user's by email (which is their username). Wordpress already does this
image

Issue #127, opened 2019-11-30
Event Page


ThaiQ:

evanugarte:
We need to improve our UI from this:
![image](https://user-images.githubusercontent.com/36345325/71455110-6db61700-2748-11ea-829a-9f6adad45614.png)

okjennifer:
ideally the events page would follow something like this: 

![image](https://user-images.githubusercontent.com/59042514/71455698-0ea5d180-274b-11ea-8517-0959803762a0.png)

PR #212, opened 2020-01-11
Abstracted Event API Functions

evanugarte:
We now can call on the Event API from functions defined in the APIFunctions directory. This saves
lines of code and abstracts axios requests.


evanugarte:
Yes. Assigning myself to #127

Issue #217, opened 2020-01-12
Add Google Cloud Bucket for Event Photos


evanugarte:
By adding photos to [a bucket](https://cloud.google.com/storage/docs/json_api/v1/buckets), we will be able to render custom photos that are paired with events. Our website will interact with the bucket on two pages, uploading them in the `EventAdmin` component and rendering them on the event page.

This will be completed after #215.

ThaiQ:
@evanugarte  Should we be using Firebase because we have it running already?

evanugarte:
This is the same thing as #164

PR #218, opened 2020-01-12
Added Image URL to Event Model

### LINKED ISSUE #215: Add Optional Photo URL Field to Event Model
This will allow us to render images that we can upload to a bucket on Google Cloud. If the field is left empty, we will render the default SCE logo.

evanugarte:
We can now save image links to an event. This will be implemented with #217.

Resolves #215

### DIFF
--- a/api/models/Event.js
+++ b/api/models/Event.js
@@ -1,5 +1,7 @@
 const mongoose = require('mongoose')
 const Schema = mongoose.Schema
+const DEFAULT_PHOTO_URL =
+  'https://sce.engr.sjsu.edu/wp-content/uploads/2016/04/SCE_sq.png'
 
 const EventSchema = new Schema(
   {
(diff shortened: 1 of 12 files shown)


api/routes/Event.js
```diff
@@ -78,7 +77,7 @@ router.post('/editEvent', (req, res) => {
       event.startTime = startTime || event.startTime
       event.endTime = endTime || event.endTime
       event.eventCategory = eventCategory || event.eventCategory
-      // save updates
+      event.imageURL = imageURL || event.imageURL
```
ThaiQ: Will this be the URL link to fire base?

api/models/Event.js
```diff
@@ -28,6 +29,10 @@ const EventSchema = new Schema(
     },
     eventCategory: {
       type: String
+    },
+    imageURL: {
```
ThaiQ: Will this be the URL link to fire base? how will this work?

Issue #151, opened 2019-12-21
Connect React Components with APIs


evanugarte:
- Develop functions to sit in the “middle” of APIs and React components
- Similar to [this file](https://github.com/CS157A-Team-30/StoreFront_Team30/blob/master/storefrontApplication/src/APIFunctions.js) Each function will have an appropriate name and make axios requests.
- The functions will be asynchronous and be imported in React files. 
- **Benefit:** We would be able to abstract the axios requests, and we can test the functions, reducing lines of code.

PR #221, opened 2020-01-15
Added Email Verification

### LINKED ISSUE #202: Add Email Confirmation to Sign Up Page
After the user submits an application, they should be presented with a page similar to this:

image

arjohnston:
Resolves #202


api/config/config.js
```diff
@@ -26,5 +26,14 @@ module.exports = {
 
   // weak:
   // Unrestricted passwords allowed
-  passwordStrength: 'weak'
+  passwordStrength: 'weak',
```
evanugarte: is this file supposed to be here?

arjohnston: No, the config file should not be committed - and is now added to gitignore. I did not remove it from GitHub during commits because I don’t want to disrupt dev on other people’s accounts since this file is relied upon. When deploying we can generate a new hash and mailer keys + token when removing this file from public access

evanugarte: lets add the to the gitignore, the contributions right now are mostly me and thai so it shouldnt be too disruptive.

arjohnston:
The verification email has been integrated into the membership application page and a section in the Readme has been added with instructions of how to set up the mailer client. Keep in mind, this requires direct access into the sce.sjsu@gmail.com account

src/Pages/MembershipApplication/MembershipForm.js
```diff
@@ -73,6 +74,31 @@ export default function MembershipForm (props) {
       })
   }
 
+  async function sendVerificationEmail () {
```
evanugarte: To abstract axios requests, I put all API calls into `src/APIFunctions/`. Could you put this inside of a Profile.js file in the directory? See [Event.js as an example](https://github.com/SCE-Development/Core-v4/blob/dev/src/APIFunctions/Event.js)

src/Pages/MembershipApplication/MembershipForm.js
```diff
@@ -73,6 +74,31 @@ export default function MembershipForm (props) {
...
+      })
+      .then(async result => {
+        setEmailToVerified()
+      })
+      .catch(error => {
+        console.log(error)
+      })
+  }
+
+  async function setEmailToVerified () {
```
evanugarte: Also abstract this axios call. #151 describes the abstraciton effort

api/mailer/templates/verification.js
```diff
@@ -0,0 +1,23 @@
...
+      .then(hashedId => {
+        return resolve({
+          from: user,
+          to: recipient,
+          subject: 'Please verify your email address',
+          generateTextFromHTML: true,
+          html: `
+            Hi ${name},<br />
+            <p>Please verify your email by clicking below</p>
+            <a href='https://sce.engr.sjsu.edu/verify?id=${hashedId}&user=${recipient}'>Verify Email</a>
```
evanugarte: Let's also create a page on the website that the link directs to, It could be something very simple, like the [ConfirmationPage component](https://github.com/SCE-Development/Core-v4/blob/dev/src/Pages/MembershipApplication/ConfrimationPage.js).

src/Pages/MembershipApplication/MembershipForm.js
arjohnston: Did this is my newest push

src/Pages/MembershipApplication/MembershipForm.js
arjohnston: Completed in the newest push

api/mailer/templates/verification.js
arjohnston: Created a verifyEmail component and added it to the /verify route

api/config/config.js
ThaiQ: ok

evanugarte:
Also I can't seem to get the verification success. Is this because we are on on localhost?
image

arjohnston:
Ah gotcha, something on the membership application page to “resend email confirmation”? That’s a good idea, and everything should be in place to make that happen.

The undefined name makes me think that the verify email link also has undefined in it, hence the trouble. Can you verify the link doesn’t have undefined in it? It should have a hashed ID and user

And lastly, the only difference of localhost (your machine) and production, is that production usually uses a process manager to keep the app running 24/7, otherwise they’re both using “localhost” and should function assuming all keys are present and valid

evanugarte:
@arjohnston Testing now. be sure to update your branch

PR #236, opened 2020-01-22
Changed Default Member Privilege to Pending

ThaiQ:
Changing the default access-level to pending (-1)

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -54,7 +54,7 @@ const UserSchema = new Schema(
 
     accessLevel: {
       type: Number,
-      default: membershipState.ADMIN
+      default: membershipState.PENDING
     },
     lastLogin: {
       type: Date,
--- a/api/routes/Event.js
+++ b/api/routes/Event.js
@@ -27,10 +27,9 @@ router.post('/createEvent', (req, res) => {
   } else if (!checkIfTokenValid(req)) {
     return res.sendStatus(UNAUTHORIZED)
   }
-  logger.log(req.body.id)
   const newEvent = new Event({
     title: req.body.title,
-    description: req.body.description,
+    description: req.body.dcheckIfTokenValidescription,
     eventLocation: req.body.eventLocation,
     eventDate: req.body.eventDate,
     startTime: req.body.startTime,
--- a/api/routes/Event.js
+++ b/api/routes/Event.js
@@ -29,7 +29,7 @@ router.post('/createEvent', (req, res) => {
   }
   const newEvent = new Event({
     title: req.body.title,
-    description: req.body.dcheckIfTokenValidescription,
+    description: req.body.description,
     eventLocation: req.body.eventLocation,
     eventDate: req.body.eventDate,
     startTime: req.body.startTime,
--- a/api/routes/user.js
+++ b/api/routes/user.js
@@ -108,7 +108,7 @@ router.post('/login', function (req, res) {
           if (isMatch && !error) {
             if (
               new Date() - user.membershipValidUntil > 0 &&
-              user.accessLevel < membershipState.ADMIN
+              user.accessLevel < membershipState.PENDING
             ) {
               return res
                 .status(UNAUTHORIZED)
--- a/src/Pages/Profile/MemberView/InfoCard.js
+++ b/src/Pages/Profile/MemberView/InfoCard.js
@@ -53,7 +53,8 @@ export default function ProfileCard (props) {
       <div id='profile-box'>
         {props.fields.map((field, ind) => (
           <h3 key={ind} id='inner-text-top'>
-            <b>{field.title}:</b> {field.value}
+            <b>{field.title}:</b>{' '}
+            <nobr dangerouslySetInnerHTML={{ __html: field.value }} />
           </h3>
         ))}
         <h3 id='inner-text-top'>
--- a/src/Pages/Profile/MemberView/InfoCard.js
+++ b/src/Pages/Profile/MemberView/InfoCard.js
@@ -54,7 +54,15 @@ export default function ProfileCard (props) {
         {props.fields.map((field, ind) => (
           <h3 key={ind} id='inner-text-top'>
             <b>{field.title}:</b>{' '}
-            <nobr dangerouslySetInnerHTML={{ __html: field.value }} />
+            {
+              <span
+                className={
+                  field.value && field.value.includes('Valid') ? 'invalid' : ''
+                }
+              >
+                {field.value}
+              </span>
+            }
           </h3>
         ))}
         <h3 id='inner-text-top'>
--- a/src/Pages/Profile/MemberView/Profile.jsx
+++ b/src/Pages/Profile/MemberView/Profile.jsx
@@ -40,7 +40,9 @@ export default class Profile extends Component {
           { title: 'Door Code', value: this.state.user.doorCode },
           { title: 'Joined Date', value: this.state.user.joinDate.slice(0, 10) },
           { title: 'Email', value: this.state.user.email },
-          { title: 'Membership Expiration', value: this.state.user.membershipValidUntil.slice(0, 10) },
+          { title: 'Membership Expiration', value: (this.props.user.accessLevel<0)? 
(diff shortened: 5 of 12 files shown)


ThaiQ:
What member see when they go to profile
image

ThaiQ:
Travis failed because default user is no longer admin - hence no access to any/most api-routing-functions

ThaiQ:
"Not Valid" is not red
image

ThaiQ:
do you have any example of mocking login?

evanugarte:
### What I did

Added a mocking class for test cases. Using `APIFunctions` is helpful as not only do we abstract axios, we can also mock the output from those requests.

The class derives itself from [this existing example](https://github.com/SCE-Development/Core-v4/blob/d3a10ddf59949ed92d147dbf303a5d08c3ff3c08/test/frontend/EventList.test.js#L30)

### How to Fix the PR
1. Open `3DPrinting.js` in `test/`
2. Open `3DPrinting.js` in `api/routes/`
3. Using console logs and [cool google searches like this](https://lmgtfy.com/?q=mocking+a+function+sinon&pp=1) find out how to make the token verify function return true for when we need it to

api/routes/Event.js
```diff
@@ -27,10 +27,9 @@ router.post('/createEvent', (req, res) => {
   } else if (!checkIfTokenValid(req)) {
     return res.sendStatus(UNAUTHORIZED)
   }
-  logger.log(req.body.id)
   const newEvent = new Event({
     title: req.body.title,
-    description: req.body.description,
+    description: req.body.dcheckIfTokenValidescription,
```
evanugarte: can we keep this as `description`?

api/routes/Event.js
```diff
@@ -27,10 +27,9 @@ router.post('/createEvent', (req, res) => {
   } else if (!checkIfTokenValid(req)) {
     return res.sendStatus(UNAUTHORIZED)
   }
-  logger.log(req.body.id)
   const newEvent = new Event({
     title: req.body.title,
```
evanugarte: fun fact, by doing

```js
const obj = {
   title
}
```

this automatically implies
```js
const obj = {
   title: title
}
```

change this `NewEvent` line by doing
```js
const {
  title, eventLocation, eventDate,
  /* all other fields here */
} = req.body
```
then do

```js
 const newEvent = new Event({
  title, eventLocation, eventDatem
   /* all other fields here */
})
```

src/Pages/Profile/MemberView/InfoCard.js
```diff
@@ -53,7 +53,8 @@ export default function ProfileCard (props) {
       <div id='profile-box'>
         {props.fields.map((field, ind) => (
           <h3 key={ind} id='inner-text-top'>
-            <b>{field.title}:</b> {field.value}
+            <b>{field.title}:</b>{' '}
+            <nobr dangerouslySetInnerHTML={{ __html: field.value }} />
```
evanugarte: why are you doing this line? i dont like injecting values into HTML. it leads to [xss](https://www.youtube.com/watch?v=L5l9lSnNMxg)

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -40,7 +40,9 @@ export default class Profile extends Component {
           { title: 'Door Code', value: this.state.user.doorCode },
           { title: 'Joined Date', value: this.state.user.joinDate.slice(0, 10) },
           { title: 'Email', value: this.state.user.email },
-          { title: 'Membership Expiration', value: this.state.user.membershipValidUntil.slice(0, 10) },
+          { title: 'Membership Expiration', value: (this.props.user.accessLevel<0)? 
+          "<font color='red'>Not Valid</font>, please see us at Engr294" :
```
evanugarte: make the whole line red instead
```suggestion
          "<font color='red'>Not Valid, please see us at Engr294</font>" :
```

test/mocks/TokenValidFunctions.js
```diff
@@ -0,0 +1,18 @@
...
+
+function restoreMock () {
+  checkifTokenValidMock.restore()
+}
+
+function resetMock () {
+  checkifTokenValidMock.reset()
+}
+
+function setReturnOfTokenValidMock (returnValue) {
```
evanugarte: could you name this something that reads more like english? e.g.

```suggestion
function tokenValidWillReturn (returnValue) {
```

test/mocks/TokenValidFunctions.js
```diff
@@ -0,0 +1,18 @@
+const TokenFunctions = require('../../util/api-utils/token-functions')
+const sinon = require('sinon')
+
+const checkifTokenValidMock = sinon.stub(TokenFunctions, 'checkIfTokenValid')
+
+function restoreMock () {
```
evanugarte: add a simple comment to each function in this file, for this one you can do 

```
/**
 * This function unstubs the checkIfTokenValid function and is called at the end of each
 * test.
 */
```

src/Pages/Profile/MemberView/InfoCard.js
evanugarte:
```html
<b>{field.title}:</b>{' '}
<span className={field.something ? 'red-text' : ''}>{field.value}</span>
```

test/mocks/TokenValidFunctions.js
evanugarte: consider doing this^

ThaiQ: done

api/models/User.js
```diff
@@ -59,7 +59,7 @@ const UserSchema = new Schema(
     // 2: Admin
     accessLevel: {
       type: Number,
-      default: 2
+      default: -1
```
evanugarte: no more hardcoded numbers. [we have enums for this](https://github.com/SCE-Development/Core-v4/blob/3c4ba959ba6a462c28bae5841594edf1ed625033/src/Enums.js#L28) so import them

api/routes/user.js
```diff
@@ -107,7 +107,7 @@ router.post('/login', function (req, res) {
           if (isMatch && !error) {
             if (
               new Date() - user.membershipValidUntil > 0 &&
-              user.accessLevel < 2
+              user.accessLevel < 0
```
evanugarte: no more hardcoded numbers. [we have enums for this](https://github.com/SCE-Development/Core-v4/blob/3c4ba959ba6a462c28bae5841594edf1ed625033/src/Enums.js#L28) so import them

src/Pages/Profile/MemberView/Profile.jsx
```diff
@@ -40,7 +40,9 @@ export default class Profile extends Component {
           { title: 'Door Code', value: this.state.user.doorCode },
           { title: 'Joined Date', value: this.state.user.joinDate.slice(0, 10) },
           { title: 'Email', value: this.state.user.email },
-          { title: 'Membership Expiration', value: this.state.user.membershipValidUntil.slice(0, 10) },
+          { title: 'Membership Expiration', value: (this.props.user.accessLevel<0)?
```
evanugarte: no more hardcoded numbers. [we have enums for this](https://github.com/SCE-Development/Core-v4/blob/3c4ba959ba6a462c28bae5841594edf1ed625033/src/Enums.js#L28) so import them

api/routes/user.js
```diff
@@ -108,7 +108,7 @@ router.post('/login', function (req, res) {
           if (isMatch && !error) {
             if (
               new Date() - user.membershipValidUntil > 0 &&
-              user.accessLevel < membershipState.ADMIN
+              user.accessLevel < membershipState.PENDING
```
evanugarte: nice. but can you explicitly say that this applies to banned people with
```suggestion
              user.accessLevel === membershipState.BANNED
```

src/Pages/Profile/MemberView/Profile.jsx
evanugarte: BRUH

ThaiQ: forgot to commit

PR #254, opened 2020-01-30
Fixes OverviewBoard's update bug + Refactored OverviewBoard

ThaiQ:
Update member-dashboard when delete user
Updated email-check when user want to change their email
Allow search for first, name, and email
Added search parameter for different membership-status

### DIFF
--- a/api/routes/user.js
+++ b/api/routes/user.js
@@ -135,7 +135,7 @@ router.post('/login', function (req, res) {
                   if (error) {
                     logger.log(
                       'Bad request while trying to reset pageCount to 0 for ' +
-                      user.email
+                        user.email
                     )
                   }
 
@@ -236,12 +236,12 @@ router.post('/users', function (req, res) {
   } else if (!checkIfTokenValid(req)) {
     return res.sendStatus(UNAUTHORIZED)
   }
-  User.find({}, function (error, result) {
-    if (error) {
+  User.find()
+    .sort({ joinDate: -1 })
+    .then(items => res.status(OK).send(items))
+    .catch(() => {
       res.status(BAD_REQUEST).send({ message: 'Bad Request.' })
-    }
-    return res.status(OK).send(result)
-  })
+    })
 })
 
 // Edit/Update a member record
@@ -251,7 +251,7 @@ router.post('/edit', (req, res) => {
   } else if (!checkIfTokenValid(req)) {
     return res.sendStatus(UNAUTHORIZED)
   }
-  const query = { email: req.body.queryEmail }
+  const query = { email: req.body.email }
   const user =
     typeof req.body.numberOfSemestersToSignUpFor === 'undefined'
       ? { ...req.body }
@@ -265,7 +265,6 @@ router.post('/edit', (req, res) => {
   delete user.numberOfSemestersToSignUpFor
 
   // Remove the auth token from the form getting edited
-  delete user.queryEmail
   delete user.token
 
   User.updateOne(query, { ...user }, function (error, result) {
@@ -276,11 +275,11 @@ router.post('/edit', (req, res) => {
     if (result.nModified < 1) {
       return res
         .status(NOT_FOUND)
-        .send({ message: `${req.body.queryEmail} not found.` })
+        .send({ message: `${query.email} not found.` })
     }
 
     return res.status(OK).send({
-      message: `${req.body.queryEmail} was updated.`,
+      message: `${query.email} was updated.`,
       membershipValidUntil: user.membershipValidUntil
     })
   })
@@ -367,7 +366,7 @@ router.post('/verify', function (req, res) {
 })
 
 // Helpers
-function testPasswordStrength(password) {
+function testPasswordStrength (password) {
   const passwordStrength = config.passwordStrength || 'strong'
   /* eslint-disable */
   const strongRegex = new RegExp(
@@ -395,7 +394,7 @@ function testPasswordStrength(password) {
   return { success: mediumRegex.test(password), message: mediumMessage }
 }
 
-function checkIfPageCountResets(lastLogin) {
+function checkIfPageCountResets (lastLogin) {
   if (!lastLogin) return false
 
   const newDate = new Date()
@@ -411,7 +410,7 @@ function checkIfPageCountResets(lastLogin) {
   return false
 }
 
-function getMemberValidationDate(numberOfSemestersToSignUpFor) {
+function getMemberValidationDate (numberOfSemestersToSignUpFor) {
   const today = new Date()
   const membershipValidationDate = new Date()
 
--- a/src/APIFunctions/User.js
+++ b/src/APIFunctions/User.js
@@ -185,7 +185,7 @@ export async function editUser (userToEdit, token) {
     .post('/api/user/edit', {
       firstName,
       lastName,
-      queryEmail: email,
+      email,
       password,
       major,
       numberOfSemestersToSignUpFor,
--- a/src/Pages/Overview/Overview.js
+++ b/src/Pages/Overview/Overview.js
@@ -2,6 +2,13 @@ import React, { Component } from 'react'
 import './Overview.css'
 import OverviewProfile from './OverviewProfile.js'
 import { getAllUsers, deleteUserByEmail } from '../../APIFunctions/User'
+import {
+  ButtonDropdown,
+  DropdownToggle,
+  DropdownMenu,
+  DropdownItem
+} from 'reactstrap'
+import { membershipState } from '../../Enums'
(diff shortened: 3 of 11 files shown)


src/Pages/Profile/admin/AdminView.jsx
```diff
@@ -38,6 +38,7 @@ export default function Editor(props) {
       lastName: lastName || user.lastName,
       middleInitial: middleInitial || user.middleInitial,
       email: email || queryEmail,
```
evanugarte: `email` should not be updated on the backend. this causes a logistic headache. change the API to send only `email` as a parameter which will find the user in a search.

api/routes/user.js
```diff
@@ -236,7 +236,7 @@ router.post('/users', function (req, res) {
   } else if (!checkIfTokenValid(req)) {
     return res.sendStatus(UNAUTHORIZED)
   }
-  User.find({}, function (error, result) {
+  User.find({}, null, { sort: { joinDate: -1 } }, function (error, result) {
```
evanugarte: my man if u wanna sort do it [like this](https://github.com/SCE-Development/Core-v4/blob/1667f5a2a93afceedafa6c70f66f3581cca9aa26/api/routes/Event.js#L20)

src/APIFunctions/User.js
```diff
@@ -184,7 +185,8 @@ export async function editUser (userToEdit, token) {
     .post('/api/user/edit', {
       firstName,
       lastName,
-      queryEmail: email,
+      queryEmail: queryEmail,
```
evanugarte:
```suggestion
      queryEmail,
```

src/APIFunctions/User.js
```diff
@@ -184,7 +185,8 @@ export async function editUser (userToEdit, token) {
...
+      email: email,
```
evanugarte:
```suggestion
      email,
```

src/Pages/Profile/admin/AdminView.jsx
ThaiQ: let make an issue for this. Its an easy fix and i don't think it should be here

ThaiQ: my latest push dealt with this

ThaiQ:
New search options and input box
image

Search options
image

On click update the entire board:
image

api/routes/user.js
```diff
@@ -240,8 +240,9 @@ router.post('/users', function (req, res) {
     if (error) {
```
evanugarte: let's just do

```js
  User.find()
    .sort({ joinDate: -1 })
    .then(users => res.status(OK).send(users))
```

src/Pages/Overview/Overview.js
```diff
@@ -2,15 +2,24 @@ import React, { Component } from 'react'
...
+  DropdownItem
+} from 'reactstrap'
+import { membershipState } from '../../Enums'
 
 export default class OverviewBoard extends Component {
   constructor (props) {
     super(props)
     this.state = {
       // All users array, update by callDatabase
       users: [],
```
evanugarte: your fixing of the update bug good, can you open a seperate PR for the search types?

ThaiQ: It a pain trying to separate them now. So I woudnt like that

api/routes/user.js
ThaiQ: Fixed. also added a .catch()

src/Pages/Overview/Overview.js
ThaiQ: Its not much, so can you look at them

ThaiQ:
I wrote my component already.
After looking at your components. My used the same technique I also have [queryType](https://github.com/SCE-Development/Core-v4/blob/da1e2643df308ccac8a84d3a84a959dc39e3155f/src/Pages/Overview/Overview.js#L21) and [queryOption](https://github.com/SCE-Development/Core-v4/blob/da1e2643df308ccac8a84d3a84a959dc39e3155f/src/Pages/Overview/Overview.js#L22)

Furthermore for improvement: this [function](https://github.com/SCE-Development/Core-v4/blob/da1e2643df308ccac8a84d3a84a959dc39e3155f/src/Pages/Overview/Overview.js#L52) allows you to search for either ```First Name, Last Name, or Email``` without having to specify it on the drop-down. In another word, in the same ```input``` you can enter any of the ```First Name, Last Name, or Email``` and they user can still be found. 

The sole-purpose of the drop down is to filter for ```Membership Types```
image
the ```input``` work normally with any of the ```First Name, Last Name, or Email```

So yeah, they are essentially the same component that you recommended me awhile back and now.

api/routes/user.js
```diff
@@ -276,11 +275,11 @@ router.post('/edit', (req, res) => {
     if (result.nModified < 1) {
       return res
         .status(NOT_FOUND)
-        .send({ message: `${req.body.queryEmail} not found.` })
+        .send({ message: `${req.body.email} not found.` })
```
evanugarte: for improved readability
```suggestion
        .send({ message: `${query.email} not found.` })
```

api/routes/user.js
```diff
@@ -276,11 +275,11 @@ router.post('/edit', (req, res) => {
...
     }
 
     return res.status(OK).send({
-      message: `${req.body.queryEmail} was updated.`,
+      message: `${req.body.email} was updated.`,
```
evanugarte:
```suggestion
      message: `${query.email} was updated.`,
```

src/Pages/Overview/Overview.js
```diff
@@ -34,15 +43,48 @@ export default class OverviewBoard extends Component {
...
+
+    const userExists = user => {
+      return (
+        user.firstName.toLowerCase().includes(value) ||
+        user.lastName.toLowerCase().includes(value) ||
+        user.email.toLowerCase().includes(value)
+      )
+    }
+
+    if (this.state.currentQueryType === 'All') {
```
evanugarte: Now would be a good time to use a switch statement. [if else vs switch](https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html#if-else_versus_switch)

```js
const { currentQueryType } = this.state

switch(currentQueryType) {
  case 'Officer':
    // do your thing
    break;
  /* etc here */
  default:
    // handle the 'All' case here
}
```

src/Pages/Overview/Overview.js
```diff
@@ -34,15 +43,48 @@ export default class OverviewBoard extends Component {
...
+    } else if (this.state.currentQueryType === 'Admin') {
+      const filtered = this.state.users.filter(
+        data => data.accessLevel === membershipState.ADMIN
+      )
+      const search = this.state.users.filter(
+        data => userExists(data) && data.accessLevel === membershipState.ADMIN
+      )
+      queryResult = search.length > 0 ? search : filtered
+    } else if (this.state.currentQueryType === 'Pending') {
+      const filtered = this.state.users.filter(
```
evanugarte: what is the difference between filtered and search? can you combine the two?

ThaiQ: ```queryResult``` holds the final array of users to be map out to the dahsboard.

```filtered``` holds all users with the specified ```queryType of "Officer, Admin, or Pending"``` Render when input is empty or when a different ```queryType``` is selected.

```search``` is ```filtered``` + ```input ``` of "First name, last name, or email" (more narrowed). Render when input is not empty.

File: src/Pages/Overview/Overview.js [hunk 11]
```diff
@@ -34,15 +43,56 @@ export default class OverviewBoard extends Component {
...
+      return (
+        user.firstName.toLowerCase().includes(value) ||
+        user.lastName.toLowerCase().includes(value) ||
+        user.email.toLowerCase().includes(value)
+      )
+    }
+
+    const { currentQueryType } = this.state
+    let filtered, search
+    switch (currentQueryType) {
```
evanugarte: to reduce the complexity of this function, let's move the `switch` statement to it's own function e.g.

```js
fiterUserByAccessLevel (accessLevel) {
  switch(accessLevel) {
    case 'Officer':
      return this.state.users.filter(
          data => data.accessLevel === membershipState.OFFICER
        )
    case 'Admin':
      // do admin here
     // other cases here
    default:
      return this.state.users
  }
}
```

then calling on it would look like
```js
let filteredUsersByLevel = fiterUserByAccessLevel(currentQueryType)
let searchResult = filteredUsersByLevel.filter(data => userExists(data))
return searchResult.length ? searchResult : filteredUsersByLevel
```

ThaiQ: done

PR #328, opened 2020-02-25
Fix table overflow bug

### LINKED ISSUE #301: User Dashboard Name-limitation
This is what happen when a user has a long name.
image

----
Fix this by **only display max 12 characters** of first and last name. Longer name will be 
```12CharacterOfFirst... 12CharacterOfLast...```

The code is located in src/pages/overview/OverviewProfile.js at [line 19](https://github.com/SCE-Development/Core-v4/blob/7a9a047c9537f698cad43ca013090922265bc863/src/Pages/Overview/OverviewProfile.js#L19)

### PR DESCRIPTION
**Pull request by aashrav**
Resolves #301

### DIFF
--- a/src/APIFunctions/Profile.js
+++ b/src/APIFunctions/Profile.js
@@ -47,3 +47,20 @@ export async function setEmailToVerified (email) {
     })
   return emailSetToVerified
 }
+
+/**
+ * Formats the first and last name by making sure the first letter of both are uppercase
+ * @param {Object} user - The object contianing all of the user data fetched from mangoDB
+ * @param {String} user.firstName - The first name of the user
+ * @param {String} user.lastName - The last name of the user
+ * @returns {String} The string of the users first and last name formated
+ */
+export function formatFirstAndLastName (user) {
+  return (
+    user.firstName[0].toUpperCase() +
+    user.firstName.slice(1, user.firstName.length) +
+    ' ' +
+    user.lastName[0].toUpperCase() +
+    user.lastName.slice(1, user.lastName.length)
+  )
+}
--- a/src/Pages/Overview/Overview.css
+++ b/src/Pages/Overview/Overview.css
@@ -1,24 +1,12 @@
-.App {
-  text-align: center;
-}
-
-.App-header {
-  background-color: #282c34;
-  min-height: 100vh;
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
-  font-size: calc(10px + 2vmin);
-  color: white;
-}
-
-.App-link {
-  color: #09d3ac;
+.name{
+  max-width: 200px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  overflow-wrap: initial;
 }
 
 .layout{
-  margin: 10px auto ;
+  margin: 10px auto;
   width:700px
 }
 
@@ -52,7 +40,6 @@ thead.td{
   word-wrap: break-word;
   padding: 12px 15px;
 }
-
 
 .content-table tbody tr{
   border-bottom: 1px solid #dddddd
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/Pages/Overview/OverviewProfile.js [hunk 1]
```diff
@@ -17,11 +17,13 @@ export default function OverviewProfile (props) {
   return (
     <tr>
       <td>
-        {props.user.firstName[0].toUpperCase() +
-          props.user.firstName.slice(1, props.user.firstName.length) +
-          ' ' +
-          props.user.lastName[0].toUpperCase() +
-          props.user.lastName.slice(1, props.user.lastName.length)}
+        <div className='name'>
+          {props.user.firstName[0].toUpperCase() +
```
evanugarte: can you move this to a function in [`src/APIFunctions/Profile.js`](https://github.com/SCE-Development/Core-v4/blob/dev/src/APIFunctions)?

I ask because the same logic for first and last name is done [here in `src/Pages/Profile/MemberView/Profile.jsx`](https://github.com/SCE-Development/Core-v4/blob/05287caeb4f09ad7487212ea6f529261364c5a87/src/Pages/Profile/MemberView/Profile.jsx#L25) too. 

You can define a funciton in the file like this

```js
export function formatFirstAndLastName(user) {
  return user.firstName[0].toUpperCase() +
            user.firstName.slice(1, user.firstName.length) +
            ' ' +
            user.lastName[0].toUpperCase() +
            user.lastName.slice(1, user.lastName.length)
```

Now, you can call on this helper by importing it at the top and the line here becomes
```jsx
{formatFirstAndLastName(props.user)}
```

aashrav: The function that you defined doesn't actually format the first and last name, so should I call it getFirstAndLastName() instead?

evanugarte: You would send in a parameter `user`, which your function does the work of capitalizing the first letters of the name.

src/Pages/Overview/OverviewProfile.js
```diff
@@ -17,11 +18,7 @@ export default function OverviewProfile (props) {
   return (
     <tr>
       <td>
-        {props.user.firstName[0].toUpperCase() +
-          props.user.firstName.slice(1, props.user.firstName.length) +
-          ' ' +
-          props.user.lastName[0].toUpperCase() +
-          props.user.lastName.slice(1, props.user.lastName.length)}
+        <div className='name'>{formatFirstAndLastName(props.user)}</div>
```
evanugarte: Awesome! Can you also replace the [code here](https://github.com/SCE-Development/Core-v4/blob/05287caeb4f09ad7487212ea6f529261364c5a87/src/Pages/Profile/MemberView/Profile.jsx#L25) with your new function call?

ThaiQ:
Travis is failing, check your test cases

src/Pages/Overview/OverviewProfile.js
```diff
@@ -17,11 +17,13 @@ export default function OverviewProfile (props) {
   return (
     <tr>
       <td>
```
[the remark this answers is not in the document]
aashrav: i need the div...when I tried to format the inside the <td> tag it didnt' work

src/Pages/Overview/Overview.css
```diff
@@ -1,7 +1,12 @@
 .App {
```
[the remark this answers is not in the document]
aashrav: can I delete .App, .App-header, .App-link. That is the default code that come in the css file. I don't believe we are using them(isn't any change when I comment it out).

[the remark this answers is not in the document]
aashrav: overview.css keeps changing back to uppercase and results in bugs

src/APIFunctions/Profile.js
```diff
@@ -47,3 +47,13 @@ export async function setEmailToVerified (email) {
     })
   return emailSetToVerified
 }
+
```
evanugarte: Add a simple comment like the other functions in this code explaining what the function does with parameters and return types

PR #332, opened 2020-02-26
Create InventoryItem Schema

### LINKED ISSUE #316: New Page for Inventory
Navigate to ```api/models``` create a new file ```InventoryItem.js```

Look at this [example](https://github.com/SCE-Development/Core-v4/blob/dcdb57fbfd875dcff5fd7c65e616d4564b5c7da7/api/models/User.js#L1)

For the schema to create the following:
```Item Name: require
Price: require
Stock: N/A
Type: require
Description: N/A
Picture: N/A 

Collection : "InventoryItem"

### PR DESCRIPTION
**Pull request by aashrav**
Resolves #316

### DIFF
--- a/api/models/InventoryItem.js
+++ b/api/models/InventoryItem.js
@@ -1,0 +1,27 @@
+const mongoose = require('mongoose')
+const Schema = mongoose.Schema
+
+const InventoryItemSchema = new Schema(
+  {
+    price: {
+      type: Number,
+      required: true
+    },
+    stock: {
+      type: Number
+    },
+    category: {
+      type: String,
+      required: true
+    },
+    description: {
+      type: String
+    },
+    picture: {
+      type: String
+    }
+  },
+  { collection: 'InventoryItem' }
+)
+
+module.exports = mongoose.model('InventoryItem', InventoryItemSchema)
--- a/src/Enums.js
+++ b/src/Enums.js
@@ -14,6 +14,12 @@ const membershipPlans = {
   YEAR: 2
 }
 
+const itemCategories = {
+  DRINK: 'DRINK',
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: api/models/InventoryItem.js [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+const InventoryItemSchema = new Schema(
+  {
+    price: {
+      type: Number,
+      required: true
+    },
+    stock: {
+      type: Number
+    },
+    category: {
```
evanugarte: do you want to define some enums that this can be a category of?

e.g.

```js
const ITEM_CATEGORIES = {
  DRINK: 'DRINK',
  CANDY: 'CANDY',
  // etc
}
```

aashrav: I put "Updtated" because I created another PR with the same name, but I pushed the wrong code. I closed that PR and created a new one...put "updated" so that people don't confused between them.

evanugarte: Because that PR was never merged, it is ok to remove this Updated. Most will not see that closed one and the `Updated` will only further confuse. Me wondering why it was there is an example of this.

ThaiQ: @evanugarte @aashrav 
should the ```stock``` be a ```String```?  Because as a ```String``` we can label it either with number like ```1,2,3,...``` or ```out of stock``` or ```check with store...```

Or should we make a new one and call it ```status```?

ThaiQ:
lets make another PR for the ```catagory```

api/models/InventoryItem.js
evanugarte: Let's make stock a number, and just handle a zero value in the frontend e.g.

```js
if (stock === 0)
  // do sum about it
```

PR #334, opened 2020-02-28
Refactor 3DPrintForm Page

### LINKED ISSUE #210: Remove Unnecessary 3D Printing Model Fields
Based on 3D Printing [Model](https://github.com/SCE-Development/Core-v4/blob/dev/api/models/PrintingForm3D.js)

- Make `email` be automatically filled out, since the user has to be signed in when they make the request
- Make `name` be automatically filled out for the same reason above
- Remove `projectContact`, we can use the `email` for this

realCalvin:
Resolves #210 
Made additional changes: moved style into css file, mapped the printing color, removed alert functions and added red error messages that appear and disappear after a few seconds.

image

### DIFF
--- a/src/Pages/3DPrinting/3DPrintForm.js
+++ b/src/Pages/3DPrinting/3DPrintForm.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import 'bootstrap/dist/css/bootstrap.min.css'
+import './3D-print-form.css'
 import {
   Button,
   Form,
@@ -19,15 +20,16 @@ export default class PrintForm3D extends React.Component {
 
     // Variables that will be send to data base
     this.state = {
-      name: '',
+      name: this.props.user.name,
       color: '',
       url: '',
       projectType: '',
-      contact: '',
+      contact: this.props.user.email,
       comment: '',
       user: this.props.user,
       isLoggedIn: this.props.authenticated,
-      fill: false
+      fill: false,
+      error: ''
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/Pages/3DPrinting/3DPrintForm.js
```diff
@@ -76,38 +78,48 @@ export default class PrintForm3D extends React.Component {
...
+    if (!fill) {
+      this.displayErrorMessage()
+    }
     return fill
   }
 
+  // Displays error message if a form input is empty
+  displayErrorMessage () {
+    document.getElementById('error-message').style.display = 'block' // Show
+    setTimeout(() => {
```
evanugarte: do we need a timeout? why not just display it until the user changes the field?

src/Pages/3DPrinting/3DPrintForm.js
```diff
@@ -76,38 +78,48 @@ export default class PrintForm3D extends React.Component {
...
+      document.getElementById('error-message').style.display = 'none' // Hide
+    }, 2000)
+  }
+
   async submitApplication () {
     if (this.checkIfFormFilled()) {
+      const { name } = this.props.user
```
evanugarte:
```suggestion
      const { name, email } = this.props.user
```

src/Pages/3DPrinting/3DPrintForm.js
```diff
@@ -76,38 +78,48 @@ export default class PrintForm3D extends React.Component {
...
+      const { email } = this.props.user
```
evanugarte:
```suggestion
      const { color, comment, contact, projectType, url, } = this.state
```

src/Pages/3DPrinting/3DPrintForm.js
```diff
@@ -76,38 +78,48 @@ export default class PrintForm3D extends React.Component {
...
       const request = {
```
evanugarte:
```suggestion
      const request = {
        name, color, comment, contact, 
        projectType, url, email, progress: 'Pending'
      }
```

src/Pages/3DPrinting/3DPrintForm.js
```diff
@@ -116,6 +128,22 @@ export default class PrintForm3D extends React.Component {
 
   render () {
     const { fill } = this.state
+    const printingColors = [
+      'Select Color',
```
ThaiQ: We should make ```Any Color``` as default

src/Pages/3DPrinting/3DPrintForm.js
realCalvin: removed the timeout

src/Pages/3DPrinting/3DPrintForm.js
realCalvin: fixed

src/Pages/3DPrinting/3DPrintForm.js
realCalvin: fixed

src/Pages/3DPrinting/3DPrintForm.js
realCalvin: fixed

src/Pages/3DPrinting/3DPrintForm.js
realCalvin: fixed. also removed code for checking if color input is valid since default value is always valid

src/Pages/3DPrinting/3DPrintForm.js
evanugarte: Next time resolve the conversation if they are fixed

PR #335, opened 2020-02-28
Add Error API

### LINKED ISSUE #294: Write ErrorLog API
## Step 1: File Creation and Imports
Create `ErrorLog.js` in `api/routes`
At the top of the file, let's import some common stuff:
```js
const express = require('express')
const router = express.Router()
```
You also need to add the filename as a string in [this array](https://github.com/SCE-Development/Core-v4/blob/fe17d7d9fe73e06d04e7bf4146af8775389bc76d/api/index.js#L1)

## Step 2: Endpoints
Seema said we needed to handle a GET request to get all events, and a POST request to add an event.
We can handle an endpoint like so
```js
router.post('/yourEndpoint here', (req, res) => {
  // handle it
})
```
Make sure to respond with something fulfilling like "successfully added!"

seemavora:
Resolves #294

### DIFF
--- a/api/routes/ErrorLog.js
+++ b/api/routes/ErrorLog.js
@@ -1,0 +1,30 @@
+const express = require('express')
+const router = express.Router()
+const ErrorLog = require('../models/ErrorLog')
+const { OK, BAD_REQUEST } = require('../constants').STATUS_CODES
+
+// first param, the api "endpoint"
+router.post('/addErrorLog', (req, res) => {
+  const newError = new ErrorLog({
+    userEmail: req.body.userEmail,
+    errorTime: req.body.errorTime,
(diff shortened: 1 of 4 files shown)


api/routes/ErrorLog.js
```diff
@@ -0,0 +1,30 @@
+const express = require('express')
+const router = express.Router()
+const ErrorLog = require('../models/ErrorLog')
+const { OK, BAD_REQUEST } = require('../constants').STATUS_CODES
+
+// first param, the api "endpoint"
```
evanugarte: remove comment

seemavora: done!

PR #337, opened 2020-03-01
Added ErrorLog API Function 

### LINKED ISSUE #338: Add ErrorLog API Function
Resolves #338

seemavora:
Resolves #338

### DIFF
--- a/src/APIFunctions/ErrorLog.js
+++ b/src/APIFunctions/ErrorLog.js
@@ -1,0 +1,36 @@
+import axios from 'axios'
+
+/**
+ * Retrieve all errors.
+ * @returns {Object[]} an array of all errors
+ */
+export async function getAllErrorLogs() {
+  let allErrorLogs = []
+  await axios
+    .get('api/ErrorLog/getErrorLogs')
+    .then(res => {
+      allErrorLogs = res.data
+    })
+    .catch(err => {
+      allErrorLogs = err
+    })
+  return allErrorLogs
+}
+
+/**
+ * Add a new error.
+ * @param {Object} newError - The error that is to be added
+ * @param {(string|undefined)} newError.userEmail - The email of the user who has sent this error
+ * @param {(string|undefined)} newError.errorTime- The time the error occured
+ * @param {string} newError.apiEndpoint - The location of the error
+ * @param {string} newError.errordescription - The description of the error
+ */
+export async function addErrorLog(newError) {
+  let errorCreated = true
+
+  await axios.post('api/ErrorLog/addErrorLog', { ...newError })
+    .catch(() => {
+      errorCreated = false
+    })
+  return errorCreated
+}
--- a/test/ErrorLog.js
+++ b/test/ErrorLog.js
@@ -10,7 +10,6 @@ const { OK, BAD_REQUEST } = constants.STATUS_CODES
 let app = null
 
 const expect = chai.expect
-
 const tools = require('../util/testing-utils/tools.js')
 chai.should()
 chai.use(chaiHttp)


src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,33 @@
...
+    })
+    .catch(err => {
+      allErrorLogs = err
+    })
+  return allErrorLogs
+}
+
+export async function AddErrorLog (newError) {
+  let errorCreated = false
+  const errorToAdd = {
```
evanugarte: as mentioned before, `errorToAdd` is identitcal to the parameter `newError`. I would just pass `newError` to the axios call.

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,33 @@
...
+    userEmail: newError.userEmail,
+    errorTime: newError.errorTime,
+    apiEndpoint: newError.apiEndpoint,
+    errorDescription: newError.errorDescription
+  }
+  await axios
+    .post('/api/ErrorLog/addErrorLog', { ...errorToAdd })
```
evanugarte:
```suggestion
    .post('/api/ErrorLog/addErrorLog', { ...newError })
```

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,33 @@
+import axios from 'axios' // what is this??
+
+export async function getAllErrorLogs () {
```
evanugarte: add [jsdoc documentation](https://jsdoc.app/tags-param.html) to this function. An example you can take can be found here in [`src/APIFunctions/Event.js`](https://github.com/SCE-Development/Core-v4/blob/dfeabd591721a486e46c131c4ecb75c74e4e4935/src/APIFunctions/Event.js#L3)

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,33 @@
...
+    .then(res => {
+      allErrorLogs = res.data
+    })
+    .catch(err => {
+      allErrorLogs = err
+    })
+  return allErrorLogs
+}
+
+export async function AddErrorLog (newError) {
```
evanugarte: add [jsdoc documentation](https://jsdoc.app/tags-param.html) to this function. An example you can take can be found here in [`src/APIFunctions/Event.js`](https://github.com/SCE-Development/Core-v4/blob/dfeabd591721a486e46c131c4ecb75c74e4e4935/src/APIFunctions/Event.js#L21)

Also:
```suggestion
export async function addErrorLog (newError) {
```

src/APIFunctions/ErrorLog.js
seemavora: the reason I changed it to errorToAdd because it is giving me the following error if I change it to newError 
image

evanugarte: This is because you no longer use the variable and need to remove it.

src/APIFunctions/ErrorLog.js
evanugarte: Does this function retrieve all events? I thought it does something else

src/APIFunctions/ErrorLog.js
evanugarte: Take a look at [the schema you wrote](https://github.com/SCE-Development/Core-v4/blob/dev/api/models/ErrorLog.js). I would make the non required fields `string|undefined`

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,47 @@
...
+export async function addErrorLog (newError) {
+  let errorCreated = false
+  const errorToAdd = {
+    userEmail: newError.userEmail,
+    errorTime: newError.errorTime,
+    apiEndpoint: newError.apiEndpoint,
+    errorDescription: newError.errorDescription
+  }
+  await axios
+    .post('/api/ErrorLog/addErrorLog', { errorToAdd })
```
evanugarte: remove beginning slash and use `newError` here

src/APIFunctions/ErrorLog.js
evanugarte: remove `errorToAdd` as we are just gonna use the parameter

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,42 @@
...
+ * Add a new error.
+ * @param {Object} newError - The error that is to be added
+ * @param {string} newError.userEmail - The email of the user who has sent this error
+ * @param {string} newError.errorTime- The time the error occured
+ * @param {string} newError.apiEndpoint - The location of the error
+ * @param {(string|undefined)} newError.errordescription - The description of the error
+ */
+
+export async function addErrorLog (newError) {
+  let errorCreated = false
```
evanugarte: let's make this variable default to true. by doing this you can change the axios call to

```js
await axios
    .post('api/ErrorLog/addErrorLog', { newError })
    .catch(() => {
      errorCreated = false
    })
```

now, we can get rid of an entire `.then` block and only modify the `errorCreated` variable if the API call failed.

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,42 @@
...
+
+  await axios
+    .post('api/ErrorLog/addErrorLog', { newError })
```
evanugarte: remember why we must use `...`
```suggestion
    .post('api/ErrorLog/addErrorLog', { ...newError })
```

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,37 @@
...
+    .catch(err => {
+      allErrorLogs = err
+    })
+  return allErrorLogs
+}
+
+/**
+ * Add a new error.
+ * @param {Object} newError - The error that is to be added
+ * @param {string|undefined} newError.userEmail - The email of the user who has sent this error
```
ThaiQ: Are these optional param fields?

ThaiQ: Are these optional param fields? or must we have them

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {string|undefined} newError.errorTime- The time the error occured
+ * @param {string} newError.apiEndpoint - The location of the error
```
ThaiQ: also what is the format for this one? ```api/user/login``` ?

src/APIFunctions/ErrorLog.js
evanugarte: fields of the single param, which is an object

src/APIFunctions/ErrorLog.js
evanugarte: ???

src/APIFunctions/ErrorLog.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {(string)} newError.errordescription - The description of the error
+ */
+
+export async function addErrorLog (newError) {
+  let errorCreated = true
+
+  await axios.post('api/ErrorLog/addErrorLog', { ...newError }).catch(() => {
```
evanugarte: keep the newline
```suggestion
  await axios.post('api/ErrorLog/addErrorLog', { ...newError })
  .catch(() => {
```

PR #355, opened 2020-03-12
Fix bcryptjs Functionality

### LINKED ISSUE #354: New bcryptjs Dependency Blocks Login
Since #340, the new `bcryptjs` has not been functioning the same way as our old bcrypt depencency. It has to do with saving the user to our database where the password field is empty.

evanugarte:
Resolves #354. Also fixed bug regarding token generation where only the
user's first name is returned with token information.

### DIFF
--- a/api/routes/user.js
+++ b/api/routes/user.js
@@ -147,7 +147,8 @@ router.post('/login', function (req, res) {
 
             // Include fields from the User model that should be passed to the JSON Web Token (JWT)
             const userToBeSigned = {
-              name: user.firstName,
+              firstName: user.firstName,
+              lastName: user.lastName,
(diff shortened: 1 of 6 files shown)


api/routes/user.js
```diff
@@ -75,7 +75,7 @@ router.post('/register', function (req, res) {
       return res.status(BAD_REQUEST).send({ message: testPassword.message })
     }
 
-    newUser.save(function (error) {
+    newUser.save(function (error, result) {
```
jerrylee17: result isn't used though....?

evanugarte: Thanks, it is now removed.

Issue #356, opened 2020-03-12
Containerize SCE Website


evanugarte:
Using [Docker](https://www.docker.com/why-docker), let's put the frontend, server and RPC code into containers. This will allow for scalability and better follow industry standards.

PR #357, opened 2020-03-12
Downsize Dependency Count + Create util/ Directory

evanugarte:
In preparation for #356, I used npm dedupe and npm prune to lessen the size of our node_modules. I
also gave api/ it's own util directory for making containers easier to create.


evanugarte:
top 3 on the way 👿👿👿
image

PR #363, opened 2020-03-16
Add Officer Schema

### LINKED ISSUE #347: Refactor Officer Schema
Schema can be found [here](https://github.com/SCE-Development/Core-v4/blob/c4f8c7c5c77a44e281ee8ce0057580713c2007db/api/models/OfficerManager.js#L4)

Learn how to extend schema from [here](https://stackoverflow.com/questions/18317284/mongoose-extending-schemas)

Schema fields:
- quote : String
- picture : String, require, [default](https://github.com/SCE-Development/Core-v4/blob/c4f8c7c5c77a44e281ee8ce0057580713c2007db/api/constants.js#L9)
- Team: String

Create an enum var [here]()

```js
const teamType = {
  DEVELOPMENT: 'Development',
  PUBLIC_RELATIONS: 'Public Relations',
  ...
  PRESIDENT: "President"
  ...
}
```

### LINKED ISSUE #364: Refactor officer manager api
Refactor officer manager api / Fix the test case for officer manager

jeffersonly:
Resolves #347 
Resolves #364

### DIFF
--- a/api/models/OfficerManager.js
+++ b/api/models/OfficerManager.js
@@ -24,7 +24,7 @@ const OfficerSchema = new Schema(
     quote: {
       type: String
     },
-    picture: {
+    pictureUrl: {
       type: String,
       default: DEFAULT_PHOTO_URL
     }
--- a/api/models/OfficerManager.js
+++ b/api/models/OfficerManager.js
@@ -1,12 +1,9 @@
 const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
+const { DEFAULT_PHOTO_URL } = require('../constants');
 
 const OfficerSchema = new Schema(
   {
-    name: {
-      type: String,
-      required: true
-    },
     email: {
       type: String,
       required: true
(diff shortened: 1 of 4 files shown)


ThaiQ:
@evanugarte review

jeffersonly:
@evanugarte Thai mentioned that we don't need officers' names in the schema and that we could just use their email w/in the officer schema in relation to the user schema to get their name and other fields we might need

api/models/OfficerManager.js
```diff
@@ -17,29 +14,19 @@ const OfficerSchema = new Schema(
...
-    },
     team: {
       type: String,
       require: true
     },
     quote: {
       type: String
     },
-    major: {
+    picture: {
```
evanugarte: `picture` is a bit ambigious, and in the future it may improve readability if we do
```suggestion
    pictureUrl: {
```

```jsx

// pretend we are rendering an array of officers
{officers.map((officer, index) => {
  return (
    <div>
      {/* react stuff here */}
      <Image src={officer.pictureUrl} alt='officer profile' />
    </div>
  )
})}
```

PR #365, opened 2020-03-17
Store UI

cophagoknow:
make a page for the store UI


evanugarte:
This is a good start, but I don't want to merge an empty page into dev. Let's open a PR for UI when when it is ready to go.

PR #379, opened 2020-03-20
Save User's Emails as Lowercase

hsorathia:
Resolves registered user's emails not being found if they used some sort of capitalization.


hsorathia:
What error is this?

image

evanugarte:
@hsorathia just rerun travis build it goes away

PR #390, opened 2020-03-23
Refactor Admin Navbar Routing

### LINKED ISSUE #201: Add Admin Navbar Disabling to Routing.js
Adding another optional field that tell which navbar should the component have.

Before [this becomes super long](https://github.com/SCE-Development/Core-v4/blob/2a0c107a7229baf5dd97e34f66ab08ea36052042/src/Routing.js#L121)

[At here](https://github.com/SCE-Development/Core-v4/blob/2a0c107a7229baf5dd97e34f66ab08ea36052042/src/Routing.js#L33)  add an optional field to ```signedInRoutes```
```js
// ...
{
  Component: Login,
  path: '/login',
  allowedIf: !userIsAuthenticated,
  redirect: '/',
  // add the prop below
  disableAdminNavbar: true
}
```

then, the prop for NavbarWrapper can become

```js
enableAdminNavbar={!disableAdminNavbar}
```

ThaiQ:
resolves #201

### DIFF
--- a/src/Routing.js
+++ b/src/Routing.js
@@ -39,25 +39,29 @@ export default function Routing({ appProps }) {
       Component: Overview,
       path: '/dashboard',
       allowedIf: userIsOfficerOrAdmin,
-      redirect: '/'
+      redirect: '/',
+      inAdminNavbar: true
     },
     {
       Component: EventManager,
       path: '/event-manager',
       allowedIf: userIsOfficerOrAdmin,
-      redirect: '/'
+      redirect: '/',
+      inAdminNavbar: true
     },
     {
       Component: SolidsConsole,
       path: '/3DConsole',
       allowedIf: userIsOfficerOrAdmin,
-      redirect: '/'
+      redirect: '/',
+      inAdminNavbar: true
     },
     {
       Component: LedSign,
       path: '/led-sign',
       allowedIf: userIsOfficerOrAdmin,
-      redirect: '/'
+      redirect: '/',
+      inAdminNavbar: true
     },
     {
       Component: PrintingSolids,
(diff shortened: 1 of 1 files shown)


src/Routing.js
```diff
@@ -39,55 +39,64 @@ export default function Routing({ appProps }) {
...
+      redirect: '/',
+      InAdminNavbar: true
     },
     {
       Component: PrintingSolids,
       path: '/3DPrintingForm',
       allowedIf: userIsMember || userIsOfficerOrAdmin,
-      redirect: '/login'
+      redirect: '/login',
+      InAdminNavbar: false
```
evanugarte: Instead of explicitly defining them as false, you can just omit the field when you dont need it.

[on this line](https://github.com/SCE-Development/Core-v4/blob/fe1b1d56d8bffd86e5a41425bdb11fa59bf0bded/src/Components/Navbar/NavBarWrapper.js#L19) we check if we should enable the navbar or not, and it will work the same if we do
```suggestion
```

src/Routing.js
```diff
@@ -39,55 +39,64 @@ export default function Routing({ appProps }) {
       Component: Overview,
       path: '/dashboard',
       allowedIf: userIsOfficerOrAdmin,
-      redirect: '/'
+      redirect: '/',
+      InAdminNavbar: true
```
evanugarte: let's stick to using camelcase ([google suggests it](https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined))
```suggestion
      enableAdminNavbar: true
```

src/Routing.js
```diff
@@ -116,12 +125,7 @@ export default function Routing({ appProps }) {
                 component={props => (
                   <NavBarWrapper
                     component={Component}
-                    enableAdminNavbar={
-                      userIsOfficerOrAdmin &&
-                      path !== '/2DPrinting' &&
-                      path !== '/3DPrintingForm' &&
-                      path !== '/profile'
-                    }
+                    enableAdminNavbar={InAdminNavbar}
```
evanugarte:
```suggestion
                    enableAdminNavbar={enableAdminNavbar}
```

ThaiQ:
@evanugarte I am keeping ```inAdminNavbar``` because It make more sense than ```enable``` (can be miss-leading a bit)

PR #405, opened 2020-03-27
Add 3D printing api and testing

### LINKED ISSUE #269: Add RPCs to 3D Print Remotely

law-nguyen:
Resolves #269

### DIFF
--- a/test/3DPrinting.js
+++ b/test/3DPrinting.js
@@ -1,0 +1,73 @@
+process.env.NODE_ENV = 'test';
+
+const chai = require('chai');
+const chaiHttp = require('chai-http');
+const constants = require('../api/constants');
+const { OK, BAD_REQUEST } = constants.STATUS_CODES;
+const tools = require('../util/testing-utils/tools.js');
+const send3dPrintRequest =
+  require('../api/printingRPC/client/printing3d/print_3d_client');
+const sinon = require('sinon');
+
+let app = null;
+const expect = chai.expect;
+
+chai.should();
+chai.use(chaiHttp);
+const SUCCESS_MESSAGE = 'success';
+const ERROR_MESSAGE = 'error';
+const TEXT_REQUEST = {
+  raw: '',
+  name: 'cool guy evan',
+  volume: 1,
+  copies: 1
+};
+
+describe('3DPrinting', () => {
+  const send3DPrintRequestMock = sinon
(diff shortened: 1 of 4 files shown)


law-nguyen:
not sure why there are changes in the .json file :ooo

test/3DPrinting.js
```diff
@@ -0,0 +1,73 @@
+process.env.NODE_ENV = 'test';
+
+const chai = require('chai');
+const chaiHttp = require('chai-http');
+const constants = require('../api/constants');
+const { OK, BAD_REQUEST } = constants.STATUS_CODES;
+const tools = require('../util/testing-utils/tools.js');
+const send3dPrintRequest
```
evanugarte, line 8: can you put the equals sign on the same line like this
```js
const send3dPrintRequest = 
     require('../api/printingRPC/client/printing3d/print_3d_client');
```

test/3DPrinting.js
```diff
@@ -0,0 +1,73 @@
...
+
+let app = null;
+const expect = chai.expect;
+
+chai.should();
+chai.use(chaiHttp);
+const SUCCESS_MESSAGE = 'success';
+const ERROR_MESSAGE = 'error';
+const TEXT_REQUEST = {
+  raw: '',
```
evanugarte, line 20: why not put something as a value here?

test/3DPrinting.js
```diff
@@ -0,0 +1,73 @@
...
+        .send(TEXT_REQUEST)
+        .then(function(res) {
+          expect(res).to.have.status(OK);
+          done();
+        })
+        .catch(err => {
+          throw err;
+        });
+    });
+    it('Should return statuscode 400 for invalid requests', done => {
```
evanugarte, line 59:
```suggestion
    it('Should return statuscode 400 if the RPC fails', done => {
```

test/3DPrinting.js
law-nguyen: I based it off of jerry's test file, he didn't put anything there. Should I change it?

jerrylee17: Lawrence is correct. We don't need a value. I'm assuming the raw values of 3D printing files will be massive. To pass in the raw data of a test print would take up too much space. We are testing the ability for the RPC to be executed, not whether it can print a file or not. Thus, an empty string would be fine.

PR #416, opened 2020-04-01
Store Front End

### LINKED ISSUE #278: Creating the "Store" page
Navigate to ```src/Pages``` make a new folder calls ```Store``` and navigate into it.
In Store, create ```Store.js```

In Store.js, to initiate React and its basic structures do:
```js
import React from 'react'
export default class Store extends React.Component {
  constructor (props) {
    super(props)

    // Variables
    this.state = {
         yourVariable: "Go In HERE!"
    }

    render() {
         return <div>PAGE CONTENT</div>
    }
  }
}
```
---------------------------------------------------------------------
Go to src/Routing.js

1. Below line [23](https://github.com/SCE-Development/Core-v4/blob/4f16a40b171ce0addea42dc00bfb966a09e1388a/src/Routing.js#L23) import your ```Store.js```
2. At the end of [this list](https://github.com/SCE-Development/Core-v4/blob/4f16a40b171ce0addea42dc00bfb966a09e1388a/src/Routing.js#L92) add your ```Store``` which you imported  with the ```path:'/store'``` 

----------------------------------------------------------------------
You can now navigate to your page with ```localhost:3000/store```

jerrylee17:
`const example` is an example of an object that an API function in the back end will return. 
Also made Store a route that can be accessible by only members in routing. 

Resolves #278


evanugarte:
Closing this until #417 is completed

PR #431, opened 2020-04-10
Move /register contents into a util file

hsorathia:


api/util/registerUser.js
```diff
@@ -0,0 +1,130 @@
...
+    status: 'OK'
+  };
+  if (req.email && req.password) {
+    const newUser = new User({
+      password: req.password,
+      firstName: req.firstName,
+      middleInitial: req.middleInitial || '',
+      lastName: req.lastName,
+      email: req.email.toLowerCase(),
+      major: req.major || ''
```
seemavora, line 25: I think that u can just leave it as password, firstName, middleInitial, etc. The req. '' isn't required.

evanugarte: yes! can you change the parameters of the function?

hsorathia: Because i'm planning on using this register function for the terminal app to create a user as well, i'm passing in a general `req` object to the function and not individual parameters like `firstName`, `lastName`, `email`, etc.. 
So, it doesn't feel right to add more parameters because eventually I need to also manipulate the User's access level and their email verified value from the terminal application. If I do it this way I would need to add another two parameters to the function, and add two dummy values to pass in from the routes end. I can still change it and do it differently as to how I have it but I don't see the need to atm, do you still want me to do it?

hsorathia: image

hsorathia: Alt Text

api/util/registerUser.js
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Register a new user.
+ * @param {Object} newUser - The user that is to be registered
+ * @param {(string|undefined)} - The information of the User
+ *                               that is to be added
+ */
+
+
+async function registerUser(req){
```
evanugarte, line 12: can you make the variable name more descriptive
```suggestion
async function registerUser(userToAdd){
```

api/util/registerUser.js
```diff
@@ -0,0 +1,130 @@
+const User = require('../models/User');
+const config = require('../config/config');
+
+/**
+ * Register a new user.
+ * @param {Object} newUser - The user that is to be registered
+ * @param {(string|undefined)} - The information of the User
+ *                               that is to be added
+ */
+
```
evanugarte, line 10: remove this white line and the one below it

api/routes/user.js
```diff
@@ -367,35 +349,7 @@ router.post('/verify', function(req, res) {
...
-  // allow unrestricted passwords if strength is set to weak
-  if (passwordStrength === 'weak') {
-    return { success: true, message: '' };
-  }
-
-  // test medium password by default
-  return { success: mediumRegex.test(password), message: mediumMessage };
-}
-
+// Helper
```
evanugarte, line 352: remove this line

api/routes/user.js
```diff
@@ -53,35 +54,16 @@ router.post('/checkIfUserExists', (req, res) => {
...
-      req.body.numberOfSemestersToSignUpFor
-    );
-    newUser.membershipValidUntil = membershipValidUntil;
-
-    const testPassword = testPasswordStrength(req.body.password);
-
-    if (!testPassword.success) {
-      return res.status(BAD_REQUEST).send({ message: testPassword.message });
+router.post('/register', async (req, res) => {
+  const regRes = await registerUser(req.body);
```
evanugarte, line 58: Name this something better, e.g.
```suggestion
  const registrationStatus = await registerUser(req.body);
```

api/util/registerUser.js
```diff
@@ -0,0 +1,127 @@
+const User = require('../models/User');
+const config = require('../config/config');
+
+/**
+ * Register a new user.
+ * @param {Object} newUser - The user that is to be registered
+ * @param {(Object} userToAdd - The information about the user that is
+ *                              registering
+ */
```
evanugarte, line 9: i think this function returns something, so document it
```suggestion
 * @returns {Object} blah blah blah
 */
```

api/util/registerUser.js
```diff
@@ -0,0 +1,127 @@
+const User = require('../models/User');
+const config = require('../config/config');
+
+/**
+ * Register a new user.
+ * @param {Object} newUser - The user that is to be registered
+ * @param {(Object} userToAdd - The information about the user that is
```
evanugarte, line 7: remove parenthesis
```suggestion
 * @param {Object} userToAdd - The information about the user that is
```

api/util/registerUser.js
```diff
@@ -0,0 +1,127 @@
...
+ *                              registering
+ */
+async function registerUser(userToAdd){
+  let res = {
```
evanugarte, line 11: name this something better
```suggestion
  let result = {
```

api/util/registerUser.js
```diff
@@ -0,0 +1,127 @@
...
+    userSaved: true,
+    message: '',
+    status: 'OK'
+  };
+  if (userToAdd.email && userToAdd.password) {
```
evanugarte, line 16: This is good to have, but there is an issue with the code. You initialize `res.userSaved` to be true by default, but what if there is no email or password? Right now, the function would just skip the `if` block and return `res`.

Either initialize `userSaved` to `false` or have an `else` block which sets it to false

api/util/registerUser.js
```diff
@@ -0,0 +1,134 @@
+const User = require('../models/User');
+const config = require('../config/config');
+
+/**
+ * Register a new user.
+ * @param {Object} newUser - The user that is to be registered
```
evanugarte, line 6: you are saying this function takes 3 params when that isnt the case, pleas fix

PR #443, opened 2020-04-14
Create a component for header

### LINKED ISSUE #434: Make headers uniform format
Make a component that will be used to render headers.

jerrylee17:
Resolves #434

To use the header, import it to your file: 
```js
import Header from
  '../../Components/Header/Header.js';
```

Create props
```js
const headerProps = {
    title: 'SCE Printing System'
};
```

Create the component
```js
<Header {...headerProps} />
```

The result is a header that looks like this:

### DIFF
--- a/src/Components/Header/Header.js
+++ b/src/Components/Header/Header.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import {
+  Jumbotron
+} from 'reactstrap';
+
+export default function Header(props) {
+  const { title } = props;
+  return (
+    <Jumbotron>
+      <div className='text-center'>
+        <h1 className='display-4'>{title}</h1>
+      </div>
+    </Jumbotron>
+  );
+}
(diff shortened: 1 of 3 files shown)


evanugarte:
@jerrylee17 can you add some screenshots on what will change and rebase?

jerrylee17:
> Few things:
> 
> 1. Please add test cases for this component
> 2. Where else would you like to use this?
> 3. Add labels to this pr, the milestone and assign yourself

1. Ok doing this now
2. Should I just add this to wherever I see fit? I think all pages should have it so the SCE website has a uniform format. 
3. Done

evanugarte:
One of the things I want to improve is our UI. While this component is helpful, this black font on grey title is not it for styling in my opinion.

evanugarte:
@jerrylee17 Lets work to improve the UI of 2D printing on it's own instead

evanugarte:
I think the intention here is good, but we should focus on improving the UI of 2D and 3D pages individually. After we can move onto working on this.

PR #444, opened 2020-04-14
Have server listen on a different port when running tests

### LINKED ISSUE #420: Have server listen on a different port when running tests
When we run API tests, we instantiate a server to handle sending and receiving requests. However, that server listens to the same port every time when we initialize it, and as a result we cannot run the website server and api tests at the same time.

### To fix
All your changes will be inside of [`api/server.js`](https://github.com/SCE-Development/Core-v4/blob/476a39838c427c0d319976327dd5fa7f76c24a6b/api/server.js)
1. Notice that we have a variable called `testEnv` that is true or false depending on if we are in a test environment [here](https://github.com/SCE-Development/Core-v4/blob/476a39838c427c0d319976327dd5fa7f76c24a6b/api/server.js#L20)
1. We assign the port we should use on [this line](https://github.com/SCE-Development/Core-v4/blob/476a39838c427c0d319976327dd5fa7f76c24a6b/api/server.js#L76)
1. Can you assign the port to `settings.port` if we aren't testing, and a different value if we are. Try using [ternary operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) for this.

nanarboursal:
Resolves #420

### DIFF
--- a/api/server.js
+++ b/api/server.js
@@ -74,7 +74,7 @@ class Server {
     // Main Server Routine - Listen for requests on specified port
     if (!port) {
       logger.log(`Using default port ${settings.port}`, handlerTag)
-      testEnv ? port = TESTING_PORT : port = settings.port 
+      port = testEnv ? TESTING_PORT : port = settings.port 
     } else {
       logger.log(`Using custom port ${port}`, handlerTag)
       settings.port = port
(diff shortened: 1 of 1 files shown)


api/server.js
```diff
@@ -73,7 +74,7 @@ class Server {
     // Main Server Routine - Listen for requests on specified port
     if (!port) {
       logger.log(`Using default port ${settings.port}`, handlerTag)
-      port = settings.port
+      testEnv ? port = TESTING_PORT : port = settings.port
```
evanugarte, line 77:
```suggestion
      port = testEnv ? TESTING_PORT : settings.port; 
```

PR #462, opened 2020-05-06
Refactored All Tests with SceApiTester

### LINKED ISSUE #423: Create class to handle API testing
When we look at any `.js` in the `test/directory`, we see so many duplicated lines:
image

To save lines and make testing quicker and easier, we could create a class called `SceApiTester`.

## Example use
Here is an example api test in [`test/Event.js`](https://github.com/SCE-Development/Core-v4/blob/c86c7d54542074ec449201dc3d50c91eee9c113c/test/Event.js#L79)
```js
describe('/POST createEvent', () => {
    it('Should return 403 when an invalid token is supplied', done => {
      chai
        .request(app)
        .post('/api/event/createEvent')
        .send(EVENT_WITH_INVALID_TOKEN)
        .then(function(res) {
          expect(res).to.have.status(UNAUTHORIZED);
          done();
        })
        .catch(err => {
          throw err;
        });
    });
});
```
Using the class, we could refactor this test to become:
```js
describe('/POST createEvent', () => {
    const eventApiTester = new SceApiTester(app);
    it('Should return 403 when an invalid token is supplied', done => {
      const apiResponse = eventApiTester.sendPostRequest('api/event/createEvent',
                                  EVENT_WITH_INVALID_TOKEN);
      expect(apiResponse).to.have.status(UNAUTHORIZED);
      done();
    });
});
```
### Class constructor
As you saw above, lets send in the parameter `app` to the constructor so we can make axios calls within the class.

An example axios call in the class will look something like this:
```js
// define the variable containing any API response
let response = null;
// make the api request
await chai
        .request(this.app) // this.app assigned in constructor
        .post(endpoint)  // function parameter
        .send(params)  // function parameter
        .then(function(res) {
          response = res; // assign API response to variable defined at the top
        })
        .catch(err => {
          throw err; // throw err as we dont wanna have errors in testing
        });
return response; // return the response back to whoever called it
```

### Helpful functions
```js
/**
  * Sends a post request to a given endpoint
  * @param string {endpoint} - the given endpoint to call
  * @param object {params} - Any parameters to send with the API call
  * @param object {response} - The response from the API
  */
function sendPostRequest(endpoint, params) {
  /* do axios post request here */
}
```
Do the same for `sendGetRequest`.

law-nguyen:
Resolves #423

### DIFF
--- a/test/InventoryItem.js
+++ b/test/InventoryItem.js
@@ -7,8 +7,10 @@ const chai = require('chai');
 const chaiHttp = require('chai-http');
 const constants = require('../api/constants');
 const { OK, BAD_REQUEST, UNAUTHORIZED, NOT_FOUND } = constants.STATUS_CODES;
+const SceApiTester = require('../test/util/tools/SceApiTester');
 
 let app = null;
+let test = null;
 
 const expect = chai.expect;
 // tools for testing
@@ -28,6 +30,7 @@ describe('InventoryItem', () => {
     initializeMock();
     app = tools.initializeServer(
       __dirname + '/../api/routes/InventoryItem.js');
+    test = new SceApiTester(app);
     // Before each test we empty the database
     tools.emptySchema(InventoryItem);
     done();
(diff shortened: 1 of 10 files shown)


test/InventoryItem.js
```diff
@@ -72,53 +75,45 @@ describe('InventoryItem', () => {
...
+      // expect(getItemsResponse).to.have.length(1);
+      // expect(getItemsResponse[0].name).to.equal(VALID_NEW_ITEM.name);
+      // expect(getItemsResponse[0].price).to.equal(VALID_NEW_ITEM.price);
+      // expect(getItemsResponse[0].stock).to.equal(VALID_NEW_ITEM.stock);
+      // expect(getItemsResponse[0].category).to
+      //   .equal(VALID_NEW_ITEM.category);
+      // expect(getItemsResponse[0].picture).to.equal(VALID_NEW_ITEM.picture);
+      // expect(getItemsResponse[0].description).to.equal(
+      //   VALID_NEW_ITEM.description);
+      // itemName = getItemsResponse[0].name;
```
law-nguyen, line 116: In the routes, this is a POST call, not a get call, so I was a bit confused.

api/routes/Event.js
```diff
@@ -40,6 +40,7 @@ router.post('/createEvent', (req, res) => {
 
   Event.create(newEvent, (error, post) => {
     if (error) {
+      // console.log(error);
```
evanugarte, line 43: remove line

test/InventoryItem.js
evanugarte: Feel free to change it to `GET`

test/LedSign.js
```diff
@@ -87,19 +90,24 @@ describe('LedSign', () => {
...
-        })
-        .catch(err => {
-          throw err;
-        });
+      const response = await test.sendPostRequest(
+        '/api/LedSign/healthCheck', officer);
+      signResponse = response.body;
+      expect(response).to.have.status(OK);
+      // chai
+      //   .request(app)
```
seemavora, line 101: remove comments before merging!

test/SceApiTester.js
```diff
@@ -0,0 +1,57 @@
...
+      .post(endpoint)
+      .send(params)
+      .then(function(res) {
+        response = res;
+      })
+      .catch(err =>{
+        throw err;
+      });
+    return response;
+  }
```
seemavora, line 26: not a huge deal but need a blank line i believe

evanugarte:
@ThaiQ me and @law-nguyen are doing that now

PR #477, opened 2020-05-22
Make header component

### LINKED ISSUE #434: Make headers uniform format

jerrylee17:
Resolves #434

To use the header, import it to your file: 
```js
import Header from
  '../../Components/Header/Header.js';
```

Create props
```js
const headerProps = {
    title: 'SCE Printing System'
};
```

Create the component
```js
<Header {...headerProps} />
```

The result is a header that looks like this:


src/Pages/Overview/Overview.js
```diff
@@ -125,99 +129,102 @@ export default class OverviewBoard extends Component {
...
-                  />
-                );
-              })}
-          </tbody>
-        </table>
+      <div>
+        <Header {...this.headerProps} />
+        <div className='layout'>
+
+
```
seemavora, line 136: remove blank space

IsitaB:
Delete the unused background image files

jerrylee17:
> Can you allow the header text to have a bit more width? it looks like this when the page becomes narrow but still has plenty of space
> image

Margin changed from 30% to 10%

PR #484, opened 2020-05-29
Add New Server to Tests & Refactor Token Mocking

evanugarte:
Sinon and mocha perform a lot of operations in the background. This new solution
allows us to decide where and when we want to mock token functions and better
isolates each api test.

## What is this PR doing?
- We are integrating the classes defined in #485 to our tests, but token mocking broke in the process. So this is a larger testing refactor.

## Why did i not split it up into smaller PRs?
- Because this new form of token mocking requires us to use the new server file, as each test now only sends one file's path to `SceHttpServer`.

## Token mocks
- We now use [`proxyquire`](https://www.npmjs.com/package/proxyquire) to mock the import of `jsonwebtoken` in [`api/util/token-functions.js`](https://github.com/SCE-Development/Core-v4/blob/dev/api/util/token-functions.js).


symbolic_full_path
```diff
@@ -0,0 +1 @@
+/Users/evan/Documents/Core-v4/util/mock-database.js
```
jerrylee17, line 1: Seems like a local file... Maybe gitignore it or use a relative path?

evanugarte: thank u for pointing this out!! removing it now.

PR #487, opened 2020-05-30
Create Mailer API Endpoints

evanugarte:
Before, sending emails was done by the user API in a very confusing way. Now the
mailing service has its own API endpoints.


src/Pages/MembershipApplication/MembershipPlan.js
```diff
@@ -162,6 +163,11 @@ class MembershipPlan extends Component {
             </div>
           </div>
         </Container>
+        <Button onClick={() => {
```
IsitaB, line 166: delete testing button before merging

evanugarte: Thanks for pointing this out, done.

PR #505, opened 2020-06-10
Convert Last Tests to Use SceApiTester

law-nguyen:
Refactored test/LedSign.js with new SceApiTester class.

### DIFF
--- a/test/DoorCode.js
+++ b/test/DoorCode.js
@@ -210,19 +210,11 @@ describe('DoorCode', () => {
         expect(result).to.have.status(NOT_FOUND);
       }
     );
-    it('Should return 200 when a code is sucessfully updated', (done) => {
+    it('Should return 200 when a code is sucessfully updated', async () => {
       setTokenStatus(true);
-      chai
-        .request(app)
-        .post('/api/DoorCode/editCode')
-        .send({ token, id: codeId, ...EDITED_CODE })
-        .then(function(res) {
-          expect(res).to.have.status(OK);
-          done();
-        })
-        .catch((err) => {
-          throw err;
-        });
+      const result = await test.sendPostRequestWithToken(
+        token, '/api/DoorCode/editCode', {id: codeId, ...EDITED_CODE}
+      );
     });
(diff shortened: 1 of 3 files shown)


test/LedSign.js
```diff
@@ -107,87 +107,50 @@ describe('LedSign', () => {
         .to.equal(VALID_SIGN_REQUEST.backgroundColor);
       expect(signResponse.textColor).to.equal(VALID_SIGN_REQUEST.textColor);
       expect(signResponse.borderColor).to.equal(VALID_SIGN_REQUEST.borderColor);
-      done();
     });
-    it('Should return statusCode 404 when the sign is down', done => {
+    it('Should return statusCode 404 when the sign is down', async () => {
       healthCheckMock.rejects(ERROR_MESSAGE);
-      chai
```
hsorathia, line 114: just curious, is the point of this refactor to get rid of chai and implement our own testing scheme

test/LedSign.js
```diff
@@ -107,87 +107,50 @@ describe('LedSign', () => {
...
-        })
-        .catch(err => {
-          throw err;
-        });
+      const response = await test.sendPostRequest(
+        '/api/LedSign/updateSignText', VALID_SIGN_REQUEST);
+      expect(response).to.have.status(OK);
     });
-    it('Should return statusCode 400 when the sign is down', done => {
+    it('Should return statusCode 400 when the sign is down', async () => {
```
hsorathia, line 127: also another dumb question but why do we replace done w/ async

test/LedSign.js
law-nguyen: the refactor decreases the number of lines used for each test. We still call chai, just in another class

test/LedSign.js
law-nguyen: async because the SceApiTester functions are async

hsorathia: 🤡 ok that makes sense

test/LedSign.js
hsorathia: ok gotchu that makes sense

Issue #509, opened 2020-06-17
Fix Homepage to No Longer Have Overlaps


seemavora:
![image](https://user-images.githubusercontent.com/29296452/84856366-58e17e80-b01b-11ea-992c-08b8d0870bf9.png)

PR #510, opened 2020-06-17
Homepage No Longer Has Benefits Section 

seemavora:
Resolves Issue #509 

No longer has overlaps! 
image


package.json
```diff
@@ -11,7 +11,7 @@
     "frontend": "react-scripts start",
     "build": "react-scripts build",
```
evanugarte, line 12: can you reset this file to be what's on dev? seems to have some unintended changes

seemavora: the 2 lines from home.js were removed

PR #513, opened 2020-06-20
Change API Calls to Use RPC Service

evanugarte:
This PR is the final step in achieving the below architecture.
image

We now no longer call on RPC client functions from the backend and instead make HTTP requests direct from the frontend to SCE-RPC


jerrylee17:
> ill explain

nice description

src/APIFunctions/2DPrinting.js
```diff
@@ -48,9 +49,10 @@ export function parseRange(pages, maxPages) {
  */
 export async function printPage(data) {
   let status = new PrintApiResponse();
-  await axios.post('/api/print/submit', data).catch(() => {
-    status.error = true;
-  });
+  await axios.post(RPC_API_URL + '/SceRpcApi/Printer/sendPrintRequest', data)
```
jerrylee17, line 52: Makes sense since most of these endpoints don't really do anything at the moment.

src/config/config.example.js
```diff
@@ -2,5 +2,6 @@ module.exports = {
   sitekey: 'super secret key',
   GENERAL_API_URL: 'http://localhost:8080',
   LOGGING_API_URL: 'http://localhost:8081',
-  MAILER_API_URL: 'http://localhost:8082'
+  MAILER_API_URL: 'http://localhost:8082',
+  RPC_API_URL: 'http://localhost:8083'
```
ThaiQ, line 6: I assume these are for dockers? Are these ports open to each other when running on the local machine (without docker)?

evanugarte: yes

PR #518, opened 2020-06-23
Optimize Travis Build Script

evanugarte:
## What Changed?
1. We no longer clone SCE-RPC as we have detatched core-v4 from it with #513 
1. We consolidated a bunch lines in the `.travis.yml` file to make it easier to read
1. We make use of `npm ci` instead of `npm install` to speed up the install step
1. we cache some `npm` stuff to save extra time with the install step above too.

## How does this improve the build?
Here are the old build times:
image

Here are the new build times:
image


jerrylee17:
Rebase before merging

PR #522, opened 2020-06-25
Move API Services into their Own Directories

evanugarte:
## What does this do?
to better match how our backend is structure the `api/` directory was reorganized. see backend structure in the photo below
image

## File structure
Below we can see each folder has its own `routes` and maybe `models` directory. Notice how `google_api/` has a special `email_templates/` directory.
```
├── config/
├── util/
├── logging_api
│   ├── routes/
│   ├── models/
│   └── server.js
├── google_api
│   ├── email_templates/
│   ├── routes/
│   └── server.js
└── main_endpoints
    ├── routes/
    ├── models/
    ├── util/
    └── server.js
```
## the `util/` and `config/` directory
there is a single `util/` directory outside of these 3 folders, which contain helpful functions used across the three folders. `config/` has keys n stuff.


api/google_api/routes/mailer.js
```diff
@@ -1,13 +1,13 @@
 const express = require('express');
 const router = express.Router();
-const { SceGoogleApiHandler } = require('./SceGoogleApiHandler');
-const verification = require('./templates/verification');
-const { validateVerificationEmail, generateHashedId } = require('./auth');
+const { SceGoogleApiHandler } = require('../util/SceGoogleApiHandler');
+const verification = require('../email_templates/verification');
+// const { validateVerificationEmail } = require('../util/auth');
```
evanugarte, line 5: commented this out as we will resolve this auth issue with @law-nguyen tomorrow

api/main_endpoints/routes/3DPrintingForm.js
```diff
@@ -3,8 +3,7 @@
 const express = require('express');
 const router = express.Router();
 const PrintingForm3D = require('../models/PrintingForm3D.js');
-const settings = require('../util/settings');
-const logger = require(`${settings.util}/logger`);
```
seemavora, line 7: just wondering, why was this removed?

api/package.json
```diff
@@ -4,7 +4,10 @@
   "description": "contains all the APIs used by the SCE website",
   "main": "SceHttpServer.js",
   "scripts": {
-    "start": "node SceHttpServer.js",
+    "start": "nodemon util/SceHttpServer.js",
```
seemavora, line 7: Are we still planning on creating one package.json file for all the different subdivisions?

evanugarte: We can, it is creates some really annoying errors. For now we can keep it like this and when #356 is finished by @SOMLpj we can give it a shot

api/main_endpoints/routes/3DPrintingForm.js
evanugarte: `logger` is a file from a long time ago which would write limited logs to a file in Core-v4. It isn't used very much and that's why i took it out

jerrylee17:
Rebase

PR #523, opened 2020-06-25
Add token generation functionality

williamtnguyen:
Functionality works, but might need some refactoring
• Assumes config file is JSON (template is in this PR)
• Only works with Gmail API at the moment
• Functions are ran from the same file they're defined (idk if this matters)

### DIFF
--- a/api/google_api/util/SceGoogleApiHandler.js
+++ b/api/google_api/util/SceGoogleApiHandler.js
@@ -3,11 +3,18 @@ const readline = require('readline');
 const { google } = require('googleapis');
 const { googleApiKeys } = require('../../config/config.json');
 const nodemailer = require('nodemailer');
+const { consoleColors } = require('../../util/constants');
 
 const {
   CLIENT_SECRET, CLIENT_ID, REDIRECT_URIS,
   USER, REFRESH_TOKEN
 } = googleApiKeys;
+
+const {
+  redColor,
+  greenColor,
+  defaultColor,
+} = consoleColors;
 
 /**
  * Handles our website's backend to interace with various Google APIs. It also
@@ -61,35 +68,41 @@ class SceGoogleApiHandler {
    * This function prompts the user to visit the OAuth playground on Google
    * Cloud's website to supply an authorization code to it. After recieving the
    * code the function writes a JSON token the token file path;
+   * @param {boolean} isDevScript toggles this functions context: true for
+   * DevOps purposes, false for API endpoints.
    */
-  getNewToken() {
-    if (!this.runningInProduction) return;
+  getNewToken(isDevScript) {
+    if (!this.runningInProduction && !isDevScript) return;
 
     const authUrl = this.oAuth2Client.generateAuthUrl({
-      // eslint-disable-next-line
+      /* eslint-disable-next-line */
       access_type: 'offline',
-      scope: this.scopes
+      scope: this.scopes,
     });
-    console.debug('Authorize this app by visiting this url:', authUrl);
+    console.debug('\nAuthorize this app by visiting this URL:\n' + authUrl);
 
     const rl = readline.createInterface({
       input: process.stdin,
-      output: process.stdout
+      output: process.stdout,
     });
 
-    rl.question('Enter the code from that page here: ', code => {
-      rl.close();
+    rl.question('\nPlease enter the \'Authorization Code\' from the URL here: ',
+      authCode => {
+        rl.close();
+        this.oAuth2Client.getToken(authCode, (err, token) => {
+          if(err) console.debug(redColor +
+            'Error generating token', err + defaultColor
+          );
+          this.oAuth2Client.setCredentials(token);
 
-      this.oAuth2Client.getToken(code, (err, token) => {
-        if (err) return console.debug('Error retrieving access token', err);
-        this.oAuth2Client.setCredentials(token);
-        // Store the token to disk for later program executions
-        fs.writeFile(this.tokenPath, JSON.stringify(token), err => {
-          if (err) return console.debug('ERR', err);
-          console.debug('Token stored to', this.tokenPath);
+          // Store the token to disk for later program executions
+          console.debug(`\nGenerating token.js file to ${this.tokenPath}...`);
+          fs.writeFileSync(this.tokenPath, JSON.stringify(token));
+          console.debug(greenColor +
+            'Done! You are ready to use GCP!' + defaultColor
+          );
         });
       });
-    });
   }
 
   /**
--- a/api/util/GenerateTokenScript.js
+++ b/api/util/GenerateTokenScript.js
@@ -1,0 +1,21 @@
+const { AuthManager } = require('../google_api/util/AuthManager');
+const { SceGoogleApiHandler } =
+  require('../google_api/util/SceGoogleApiHandler');
+
+const authManager = new AuthManager();
+const configPath = __dirname + '/../config/config.json';
+authManager.setAuthCredentials(configPath, () => {
+  const scopes = [
+    'https://mail.google.com/',
+    'https://www.googleapis.com/auth/calendar'
+  ];
+  const tokenPath = __dirname + '/../config/token.json';
+  const apiHandler = new SceGoogleApiHandler(
+    scopes,
+    tokenPath,
+  );
+
+  if (apiHandler) {
+    apiHandler.getNewToken(true);
+  }
+});
--- a/api/util/constants.js
+++ b/api/util/constants.js
@@ -34,9 +34,18 @@ const teamType = {
   PRINTING_3D_CHAIR: '3D Printing/Hacking Station Specialist'
 };
 
+const consoleColors = {
+  redColor: '\x1b[31m',
+  greenColor: '\x1b[32m',
+  blueColor: '\x1b[34m',
+  yellowColor: '\x1b[33m',
+  defaultColor: '\x1b[0m',
+};
+
 module.exports = {
   STATUS_CODES,
   DEFAULT_PHOTO_URL,
   MEMBERSHIP_STATE,
-  teamType
+  teamType,
+  consoleColors,
 };
--- a/api/google_api/routes/Calendar.js
+++ b/api/google_api/routes/Calendar.js
@@ -8,7 +8,7 @@ const {
 
 router.get('/getCalendarEvents', async (req, res) => {
   const scopes = ['https://calendar.google.com/'];
-  const pathToToken = __dirname + '/../config/token.json';
+  const pathToToken = __dirname + '/../../config/token.json';
   const apiHandler = new SceGoogleApiHandler(
     scopes, pathToToken);
   const calendarID = req.query.calendarID || 'primary';
--- a/api/google_api/routes/Mailer.js
+++ b/api/google_api/routes/Mailer.js
@@ -21,7 +21,7 @@ router.post('/sendVerificationEmail', async (req, res) => {
       apiHandler.refreshToken();
     }
   } else {
-    apiHandler.getNewToken();
+    apiHandler.getNewToken(false);
   }
 
   await verification(
--- a/api/google_api/util/AuthManager.js
+++ b/api/google_api/util/AuthManager.js
@@ -1,0 +1,94 @@
+const fs = require('fs');
+const readline = require('readline');
+const { consoleColors } = require('../../util/constants');
+
+const {
+  redColor,
+  greenColor,
+  blueColor,
+  yellowColor,
+  defaultColor
+} = consoleColors;
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+  terminal: false
+});
+
+/**
+ * Handles the gathering and writing
+ * of Google API credentials
+ */
+class AuthManager {
+  /**
+   * Prompts user for input and resolves the input or false
+   * @param {String} prompt question to get the info we need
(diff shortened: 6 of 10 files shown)


api/config/config.json
```diff
@@ -0,0 +1 @@
+{"googleApiKeys":{"CLIENT_ID": null,"CLIENT_SECRET": null,"REDIRECT_URIS":["https://developers.google.com/oauthplayground"],"USER":"sce.sjsu@gmail.com","REFRESH_TOKEN":null}}
```
evanugarte, line 1: can you make this `config.example.json` and have it replace the existing `config.example.js`?

williamtnguyen: These changes are now in PR #528. Added a `config.json` to this PR to get Travis to build but it seems to still fail

src/config/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "sitekey": "super secret key",
+  "GENERAL_API_URL": "http://localhost:8080",
+  "LOGGING_API_URL": "http://localhost:8081",
+  "MAILER_API_URL": "http://localhost:8082",
+  "RPC_API_URL": "http://localhost:8083"
+}
```
law-nguyen, line 7: I noticed that most of our `config.json` files are in gitignore and we have a `config.example.json` instead. Is there any reason why this isn't the case here?

williamtnguyen: Ah yeah that'd be a conflict. I think I left this file in this commit to see if Travis would build, but now Travis runs `setup.py` which makes a `config.json` from the `config.example.json`. Will remove this file and force push

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+      config.googleApiKeys.CLIENT_SECRET = clientSecret;
+    }
+
+    fs.writeFile(configPath, JSON.stringify(config), (err) => {
+      if(err) return console.debug('ERR', err);
+      console.debug('Successfully written to:', configPath);
+
+      generateToken(
+        config,
+        ['https://mail.google.com/'],
```
evanugarte, line 32:
```suggestion
        ['https://mail.google.com/', 'https://calendar.google.com],
```

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+ */
+const setAuthCredentials = async (configPath) => {
+  const rawConfig = fs.readFileSync(configPath);
+  console.debug('Welcome to SCE\'s Google Cloud API Token Generator!');
+
+  if(rawConfig) {
+    const clientID = await inputPrompt('Please enter the Client ID: ');
+    const clientSecret = await inputPrompt('Please enter the Client Secret: ');
+
+    const config = JSON.parse(rawConfig);
```
evanugarte, line 18: you cant modify a `const` variable
```suggestion
    let config = JSON.parse(rawConfig);
```

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+      if(ans !== '') {
+        resolve(ans);
+      } else {
+        resolve(false);
+      }
+    });
+  });
+};
+
+// Driver
```
evanugarte, line 103:
```suggestion
```

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+const { google } = require('googleapis');
+
+/**
+ * Prompts user for client ID & secret and writes them to config
+ * file containing auth credentials, then generates a token.json file for GCP
+ * @param {String} configPath the directory path of the config.json file
+ */
+const setAuthCredentials = async (configPath) => {
+  const rawConfig = fs.readFileSync(configPath);
+  console.debug('Welcome to SCE\'s Google Cloud API Token Generator!');
```
evanugarte, line 12: add another console.debug prompting the user to visit [this link](https://github.com/SCE-Development/Core-v4/wiki/Getting-Started#setting-up-mailer) if they don't know what do to

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+    const config = JSON.parse(rawConfig);
+    if(clientID) {
+      config.googleApiKeys.CLIENT_ID = clientID;
+    }
+    if(clientSecret) {
+      config.googleApiKeys.CLIENT_SECRET = clientSecret;
+    }
+
+    fs.writeFile(configPath, JSON.stringify(config), (err) => {
+      if(err) return console.debug('ERR', err);
```
evanugarte, line 27: consider adding colors to print statements [like this](https://github.com/SCE-Development/Core-v4/blob/dev/api/util/CreateUserScript.js#L9)

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+
+      generateToken(
+        config,
+        ['https://mail.google.com/'],
+        __dirname + '/../config/token.json',
+      );
+    });
+
+  } else {
+    console.debug('Config file DNE at path:', configPath);
```
evanugarte, line 38: what does this mean? better to be verbose with users when we got errors.

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+ */
+const generateToken = async (configData, apiScopes, tokenPath) => {
+  const oauth2Client = new google.auth.OAuth2(
+    configData.googleApiKeys.CLIENT_ID,
+    configData.googleApiKeys.CLIENT_SECRET,
+    configData.googleApiKeys.REDIRECT_URIS[0],
+  );
+
+  const authUrl = oauth2Client.generateAuthUrl({
+    /* eslint-disable-next-line camelcase */
```
evanugarte, line 56:
```suggestion
    /* eslint-disable-next-line */
```

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+/**
+ * Prompts user for input and resolves the input or false
+ * @param {String} prompt question to get the info we need
+ */
+const inputPrompt = (prompt) => {
```
evanugarte, line 91: nice, good idea abstracting this, i'd put this at the top of the file to improve readability

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
...
+      fs.writeFileSync(tokenPath, JSON.stringify(token));
+      console.debug('Done! You are ready to use GCP!');
+    });
+
+  } else {
+    console.debug('Error with authorization code');
+  }
+};
+
+const rl = readline.createInterface({
```
evanugarte, line 82: put this at the top of the file

williamtnguyen:
Color `const`s in `debug` statements are using `+` rather than `,` because commas cause a weird indentation

e.g. 
`console.debug(color + 'string') // no indent`
`console.debug(color, 'string') // indent`

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,127 @@
+const fs = require('fs');
+const readline = require('readline');
+const { google } = require('googleapis');
+
+const redColor = '\x1b[31m';
```
evanugarte, line 5: perhaps put these color definitions in [`api/util/constants.js`](https://github.com/SCE-Development/Core-v4/blob/dev/api/util/constants.js)? then we can just import them in this script and in the create-user script

api/util/GenerateTokenScript.js
evanugarte: Add a better error message here, e.g.
```suggestion
      if(err) {
        return console.debug(`A problem occurred trying to write to ${configPath}`, err);
      }
```

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+};
+
+/**
+ * Generates an auth token to make API calls, assuming configData exists
+ * @param {Object} configData data from config file containing auth credentials
+ * @param {String | String[]} apiScopes which API(s) to interact with
+ * @param {String} tokenPath path that the auth token file will be written to
+ */
+const generateToken = async (configData, apiScopes, tokenPath) => {
```
evanugarte, line 89: is there a way you can incorporate this function into [`api/google_api/util/SceGoogleApiHandler.js`](https://github.com/SCE-Development/Core-v4/blob/dev/api/google_api/util/SceGoogleApiHandler.js)? it would be good to isolate the logic into that function.

api/google_api/util/SceGoogleApiHandler.js
```diff
@@ -61,35 +68,41 @@ class SceGoogleApiHandler {
    * This function prompts the user to visit the OAuth playground on Google
    * Cloud's website to supply an authorization code to it. After recieving the
    * code the function writes a JSON token the token file path;
+   * @param {boolean} isDevScript toggles this functions context: true for
+   * DevOps purposes, false for API endpoints.
    */
-  getNewToken() {
-    if (!this.runningInProduction) return;
+  getNewToken(isDevScript) {
```
evanugarte, line 74: i would just generate the token every time, so remove this param

api/google_api/util/SceGoogleApiHandler.js
```diff
@@ -3,12 +3,19 @@ const readline = require('readline');
 const { google } = require('googleapis');
 const { googleApiKeys } = require('../../config/config.json');
 const nodemailer = require('nodemailer');
+const { consoleColors } = require('../../util/constants');
 
 const {
   CLIENT_SECRET, CLIENT_ID, REDIRECT_URIS,
   USER, REFRESH_TOKEN
 } = googleApiKeys;
```
evanugarte, line 12:
```suggestion
```

ThaiQ:
travis failed, build error

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+  } else {
+    console.debug(redColor +
+      'config.json file does not exist at path:', configPath
+    );
+  }
+};
+
+setAuthCredentials(__dirname + '/../config/config.json');
```
ThaiQ, line 102: add an empty line at the end

api/util/constants.js
```diff
@@ -34,9 +34,18 @@ const teamType = {
...
+};
+
 module.exports = {
   STATUS_CODES,
   DEFAULT_PHOTO_URL,
   MEMBERSHIP_STATE,
-  teamType
+  teamType,
+  consoleColors,
 };
```
ThaiQ, line 51: add an empty line at the end

api/google_api/util/SceGoogleApiHandler.js
williamtnguyen: Update: the purpose of the `isDevScript` parameter is because we want this script to run without having to have the server running. The script uses `getNewToken()`, which checks `if(!this.runningInProduction) return;`, so if we ran the script without the server running, it wouldn't work. The easiest fix is to add a boolean parameter and toggle it to true so that it would bypass that initial check and run successfully without needing the server running

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,104 @@
+const fs = require('fs');
+const readline = require('readline');
+const { google } = require('googleapis');
+
+/**
```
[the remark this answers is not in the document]
williamtnguyen: Will refactor this tonight

api/util/constants.js
```diff
@@ -34,9 +34,18 @@ const teamType = {
...
+  defaultColor: '\x1b[0m',
+};
+
 module.exports = {
   STATUS_CODES,
   DEFAULT_PHOTO_URL,
   MEMBERSHIP_STATE,
-  teamType
+  teamType,
+  consoleColors,
```
hsorathia, line 50: is it standard for us to have the last item in our objects to have a `,` at the end?

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,102 @@
...
+      );
+
+      if (apiHandler) {
+        apiHandler.getNewToken(true);
+      }
+    });
+
+  } else {
+    console.debug(redColor +
+      'config.json file does not exist at path:', configPath
```
hsorathia, line 97: might need to reset the color back to `defaultColor` of the console after `configPath` so the users terminal color stays red

api/util/GenerateTokenScript.js
```diff
@@ -0,0 +1,102 @@
...
+      config.googleApiKeys.CLIENT_ID = clientID;
+    }
+    if (clientSecret) {
+      config.googleApiKeys.CLIENT_SECRET = clientSecret;
+    }
+
+    fs.writeFile(configPath, JSON.stringify(config), (err) => {
+      if (err) {
+        return console.debug(
+          `A problem occurred trying to write to ${configPath}`, err
```
hsorathia, line 73: rename `err` to `error` just so it's inline with variable names in other scripts and more readable overall

williamtnguyen:
## 2nd Refactor:
• move `setAuthCredentials()` to `AuthManager.js`, it's only purpose is to get and write google api keys
• modularize `GenerateTokenScript` to outsource work to `AuthManager` to get API keys, and `SceGoogleApiHandler` to generate token

api/util/constants.js
evanugarte: we don't have a preference as of now

PR #528, opened 2020-06-28
Refactor api & src subdirectories to use config.json

williamtnguyen:


.gitignore
```diff
@@ -81,4 +81,5 @@ build
 printingRPC/
 # Config files
 config.js
```
evanugarte, line 83: remove this line as we arent using it anymore

api/config/config.example.json
```diff
@@ -0,0 +1,19 @@
...
+  },
+  "googleCalendarIds": {
+    "PRIMARY_CAL": "primary",
+    "SCE_EVENTS_CAL": "XXXXXXXXXXXXXXXXXXXXXXXXXXX"
+  },
+  "ledSignIp": "192.168.4.1",
+  "GENERAL_API_URL": "http://localhost:8080/",
+  "LOGGING_API_URL": "http://localhost:8081/",
+  "MAILER_API_URL": "http://localhost:8081/"
+}
```
evanugarte, line 19: add blank line at bottom of this file

src/APIFunctions/2DPrinting.js
```diff
@@ -1,6 +1,6 @@
 import axios from 'axios';
 import { PrintApiResponse } from './ApiResponses';
-import { RPC_API_URL } from '../config/config';
+const { RPC_API_URL } = require('../config/config.json');
```
evanugarte, line 3: for importing stuff in `src/` use the convention like the line above:

```suggestion
import { RPC_API_URL } from '../config/config.json';
```
please fix for all other files in `src`

williamtnguyen:
This PR seems to not build whereas the other PR builds because the `config.json`'s in this PR are .gitignore'd

evanugarte:
@williamtnguyen the issue here is that when travis builds, `config.json` doesn't exist.

There are two options to fix this:
1. Create a base64 encoded environment variable for each config file and decode it to both `api/config/config.json` and `src/config/config.json`.
2. Run `setup.py` in the `install` section in `.travis.yml` like:

```yml
install:
  python setup.py
  # rest of stuff here ...
```
Note that you should remove the `setup_status` line for each os, as it clones SCE-RPC but is no longer required for tests to pass.
image

jerrylee17:
Rebase

api/config/config.example.json
```diff
@@ -0,0 +1,19 @@
...
+    "REDIRECT_URIS": ["https://developers.google.com/oauthplayground"],
+    "USER": "sce.sjsu@gmail.com",
+    "REFRESH_TOKEN": "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+  },
+  "googleCalendarIds": {
+    "PRIMARY_CAL": "primary",
+    "SCE_EVENTS_CAL": "XXXXXXXXXXXXXXXXXXXXXXXXXXX"
+  },
+  "ledSignIp": "192.168.4.1",
+  "GENERAL_API_URL": "http://localhost:8080/",
```
evanugarte, line 16: remove the slashes at the end of each URL in this file

api/config/config.example.json
```diff
@@ -0,0 +1,19 @@
...
+  "LOGGING_API_URL": "http://localhost:8081/",
+  "MAILER_API_URL": "http://localhost:8081/"
```
evanugarte, line 18: this should be 8082

PR #529, opened 2020-06-30
Added DoorCode backend

SlicerBX:
Created schema and routing for DoorCode API and wrote test cases for the routing

### DIFF
--- a/api/main_endpoints/models/DoorCode.js
+++ b/api/main_endpoints/models/DoorCode.js
@@ -5,6 +5,7 @@ const DoorCodeSchema = new Schema(
   {
     doorCode: {
       type: String,
+      unique: true,
       required: true,
     },
     doorCodeValidUntil: {
@@ -18,5 +19,14 @@ const DoorCodeSchema = new Schema(
   },
   { collection: 'DoorCodes' }
 );
+
+DoorCodeSchema.pre('save', function(next) {
+  let doorcodeRegExp = new RegExp (['^[0-9]{3}-[0-9]{4}']);
+  if (!this.doorCode.match(doorcodeRegExp)) {
+    return next('Bad door code saved (format is: xxx-xxxx)');
+  } else {
+    return next();
+  }
+});
 
 module.exports = mongoose.model('DoorCodes', DoorCodeSchema);
--- a/api/main_endpoints/routes/DoorCode.js
+++ b/api/main_endpoints/routes/DoorCode.js
@@ -1,0 +1,118 @@
+const express = require('express');
+const router = express.Router();
+const DoorCode = require('../models/DoorCode');
+const {
+  checkIfTokenSent,
+  checkIfTokenValid,
+} = require('../util/token-functions');
+const {
+  OK,
+  BAD_REQUEST,
+  UNAUTHORIZED,
+  FORBIDDEN,
+  NOT_FOUND,
+} = require('../../util/constants').STATUS_CODES;
+const addErrorLog = require ('../util/logging-helpers');
+
+router.get('/getDoorCodes', (req, res) => {
+  DoorCode.find()
+    .then(doorcodes => res.status(OK).send(doorcodes))
+    .catch(error => {
+      const info = {
+        errorTime: new Date(),
+        apiEndpoint: 'DoorCode/getDoorCodes',
+        errorDescription: error
+      };
+      addErrorLog(info);
+      res.status(BAD_REQUEST).send({ error, message: 'Getting codes failed' });
+    });
+});
+
+router.get('/getAvailableDoorCode', (req, res) => {
+  DoorCode.findOne({ usersAssigned: { $lt: 2 } }).then((doorcodes) =>
+    res.status(OK).send(doorcodes)
+  ).catch(() => {
+    res.status(BAD_REQUEST).send({ message: 'No codes left.' });
+  });
+});
+
+router.post('/addCode', (req, res) => {
+  if (!checkIfTokenSent(req)) {
+    return res.sendStatus(FORBIDDEN);
+  } else if (!checkIfTokenValid(req)) {
+    return res.sendStatus(UNAUTHORIZED);
+  }
+
+  const newCode = new DoorCode({
+    doorCode: req.body.doorCode,
+    doorCodeValidUntil: req.body.doorCodeValidUntil,
+    usersAssigned: req.body.usersAssigned,
+  });
+
+  DoorCode.create(newCode, (error, post) => {
+    if (error) {
+      return res.sendStatus(BAD_REQUEST);
+    }
+    return res.status(OK).send(post);
+  });
+});
+
+router.post('/editCode', (req, res) => {
+  if (!checkIfTokenSent(req)) {
+    return res.sendStatus(FORBIDDEN);
+  } else if (!checkIfTokenValid(req)) {
+    return res.sendStatus(UNAUTHORIZED);
+  }
+
+  const {
+    doorCode,
+    doorCodeValidUntil,
+    usersAssigned,
+  } = req.body;
+  DoorCode.findOne({ _id: req.body.id  })
+    .then((code) => {
+      code.doorCode = doorCode || code.doorCode;
+      code.doorCodeValidUntil = doorCodeValidUntil || code.doorCodeValidUntil;
+      code.usersAssigned = usersAssigned || code.usersAssigned;
+      code
+        .save()
+        .then((ret) => {
+          res.sendStatus(OK);
+        })
+        .catch((error) => {
+          res.status(BAD_REQUEST).send({
+            error,
+            message: 'door code was not updated',
+          });
+        });
+    })
+    .catch((error) => {
+      res.status(NOT_FOUND).send({ error, message: 'door code not found' });
+    });
+});
+
+router.post('/removeCode', (req, res) => {
+  if (!checkIfTokenSent(req)) {
+    return res.sendStatus(FORBIDDEN);
+  } else if (!checkIfTokenValid(req)) {
+    return res.sendStatus(UNAUTHORIZED);
+  }
+
+  DoorCode.deleteOne({ _id: req.body.id  }, (error, form) => {
+    if (error) {
+      const info = {
+        errorTime: new Date(),
+        apiEndpoint: 'DoorCode/removeItem',
+        errorDescription: error,
+      };
+      return res.sendStatus(BAD_REQUEST);
+    }
+    if (form.n < 1) {
+      return res.sendStatus(NOT_FOUND);
+    } else {
+      return res.sendStatus(OK);
+    }
+  });
+});
+
+module.exports = router;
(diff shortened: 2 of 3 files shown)


api/routes/DoorCode.js
```diff
@@ -0,0 +1,109 @@
...
+  OK,
+  BAD_REQUEST,
+  UNAUTHORIZED,
+  FORBIDDEN,
+  NOT_FOUND,
+} = require('../constants').STATUS_CODES;
+const addErrorLog = require('../util/logging-helpers');
+
+router.get('/getAvailableCodes', (req, res) => {
+  DoorCode.find({ usersAssigned: { $lt: 2 } }).then((doorcodes) =>
```
evanugarte, line 20: nice!!!

test/DoorCode.js
```diff
@@ -0,0 +1,217 @@
...
+        token,
+        '/api/DoorCode/addCode',
+        VALID_NEW_CODE
+      );
+      expect(result).to.have.status(OK);
+    });
+  });
+
+  describe('/GET getAvailableCodes', () => {
+    it('Should return an object of all codes', async () => {
```
evanugarte, line 101: can we add a test to ensure it doesnt return a code with 2 or more assignees?

api/routes/DoorCode.js
```diff
@@ -0,0 +1,109 @@
+const express = require('express');
+const router = express.Router();
+const DoorCode = require('../models/DoorCode');
+const settings = require('../util/settings');
+const logger = require(`${settings.util}/logger`);
```
evanugarte, line 5: we are about to delete `logger` and `settings`, i would remove the usage from this file

api/models/DoorCode.js
```diff
@@ -3,6 +3,11 @@ const Schema = mongoose.Schema;
 
 const DoorCodeSchema = new Schema(
   {
+    dateLastUsed: {
```
evanugarte, line 6: i am a little confused as to why we need this? let me know

api/routes/DoorCode.js
```diff
@@ -0,0 +1,109 @@
...
+});
+
+router.post('/removeCode', (req, res) => {
+  if (!checkIfTokenSent(req)) {
+    return res.sendStatus(FORBIDDEN);
+  } else if (!checkIfTokenValid(req)) {
+    return res.sendStatus(UNAUTHORIZED);
+  }
+
+  DoorCode.deleteOne({ doorCode: req.body.doorCode }, (error, form) => {
```
evanugarte, line 91: i would instead search by `_id` here, imagine the situation where we have a duplicate code, one with assignees and one without, deleting one may remove members' door codes unintentionally
```suggestion
  DoorCode.findOne({ doorCode: req.body._id })
```

for how to remove with `_id` instead check out this function in [`api/routes/Event.js`](https://github.com/SCE-Development/Core-v4/blob/ea3b2c59b57957225894fea89b1db2d24fbd10c6/api/routes/Event.js#L103) and the associated test case in [`test/Event.js`](https://github.com/SCE-Development/Core-v4/blob/ea3b2c59b57957225894fea89b1db2d24fbd10c6/test/Event.js#L201)

api/routes/DoorCode.js
```diff
@@ -0,0 +1,109 @@
...
+    return res.sendStatus(UNAUTHORIZED);
+  }
+
+  const {
+    dateLastUsed,
+    doorCode,
+    doorCodeValidUntil,
+    usersAssigned,
+  } = req.body;
+  DoorCode.findOne({ doorCode: req.body.oldDoorCode })
```
evanugarte, line 61: i would also search by `_id` here for the situation where we have a duplicate code
```suggestion
  DoorCode.findOne({ doorCode: req.body._id })
```

for how to remove with `_id` instead check out this function in [`api/routes/Event.js`](https://github.com/SCE-Development/Core-v4/blob/ea3b2c59b57957225894fea89b1db2d24fbd10c6/api/routes/Event.js#L76)

for the test case see my comment for delete and follow the same pattern

api/models/DoorCode.js
```diff
@@ -19,4 +19,5 @@ const DoorCodeSchema = new Schema(
   { collection: 'DoorCodes' }
 );
 
+// end here
```
evanugarte, line 22:
```suggestion
```

api/routes/DoorCode.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (form.n < 1) {
+      return res.sendStatus(NOT_FOUND);
+    } else {
+      return res.status(OK).send({ message: `${req.body.name} was deleted.` });
+    }
+  });
+});
+
+// end here
```
evanugarte, line 119:
```suggestion
```

api/routes/DoorCode.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  const newCode = new DoorCode({
+    doorCode: req.body.doorCode,
+    doorCodeValidUntil: req.body.doorCodeValidUntil,
+    usersAssigned: req.body.usersAssigned,
+  });
+
+  // Look for any repeated doorcodes
+  DoorCode.findOne(
```
aashrav, line 37: Instead of looking for repeated doorcodes inside the code, I realized you can add a unique field inside the schema itself. If you look in the User schema you can see that inside the email attribute they put a unique: true field. I believe this would be better and reduce code

aashrav: Also rebase after you make the changes

api/main_endpoints/routes/DoorCode.js
```diff
@@ -0,0 +1,102 @@
...
+const {
+  OK,
+  BAD_REQUEST,
+  UNAUTHORIZED,
+  FORBIDDEN,
+  NOT_FOUND,
+} = require('../../util/constants').STATUS_CODES;
+const addErrorLog = require('../util/logging-helpers');
+
+router.get('/getAvailableCodes', (req, res) => {
```
evanugarte, line 17: when would we use this? what do you think about having 2 endpoints like:

```js
// fetches all door codes, regardless of use
router.get('/getDoorCodes')

// fetches a single door code that is available so we can assign it
router.get('/getAvailableDoorCode')
```

api/main_endpoints/routes/DoorCode.js
```diff
@@ -0,0 +1,102 @@
...
+
+  DoorCode.create(newCode, (error, post) => {
+    if (error) {
+      return res.sendStatus(BAD_REQUEST);
+    }
+    return res.status(OK).send(post);
+  });
+});
+
+router.post('/editCode', (req, res) => {
```
evanugarte, line 44: would we call on this endpoint for when we are assigning a user to a doorcode?

api/main_endpoints/routes/DoorCode.js
```diff
@@ -0,0 +1,102 @@
...
+  } = req.body;
+  DoorCode.findOne({ _id: req.body.id  })
+    .then((code) => {
+      code.doorCode = doorCode || code.doorCode;
+      code.doorCodeValidUntil = doorCodeValidUntil || code.doorCodeValidUntil;
+      code.usersAssigned = usersAssigned || code.usersAssigned;
+      code
+        .save()
+        .then((ret) => {
+          res.status(OK).json({ ret, code: 'door code updated successfully' });
```
evanugarte, line 64:
```suggestion
          res.status(OK);
```

api/main_endpoints/routes/DoorCode.js
```diff
@@ -0,0 +1,102 @@
...
+        errorTime: new Date(),
+        apiEndpoint: 'DoorCode/removeItem',
+        errorDescription: error,
+      };
+      return res.sendStatus(BAD_REQUEST);
+    }
+    if (form.n < 1) {
+      return res.sendStatus(NOT_FOUND);
+    } else {
+      return res.status(OK).send({ message: `${req.body.name} was deleted.` });
```
evanugarte, line 97:
```suggestion
      return res.sendStatus(OK);
```

api/main_endpoints/models/DoorCode.js
```diff
@@ -5,6 +5,7 @@ const DoorCodeSchema = new Schema(
   {
     doorCode: {
```
evanugarte, line 6: we should add a check to ensure the doorcode follows a format of `xxx-xxxx`.

Check out mongoose's `pre('save')` function ([here](https://mongoosejs.com/docs/middleware.html#pre)) to add a check that the string is correct. an example where we use the `pre` hook is in [`api/main_endpoints/models/User.js`](https://github.com/SCE-Development/Core-v4/blob/be6648edd785a45e9fd2eac2037b3218faca9ffd/api/main_endpoints/models/User.js#L70)

check 2 things:
1. the format has three characters, a dash, then 4 characters
1. all characters beides the dash are numbers

i highly reccomend using regular expressions for this ([example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test))

api/main_endpoints/routes/DoorCode.js
SlicerBX: the editCode endpoint is so that we can edit the usersAssigned if needed and this routing was created in case the attributes of a door code needed to be changed. Do you think it should be removed?

SlicerBX:
I'd say we wait until it is merged in since we are not in a rush, but there are merits in making the limit variable right now. I would like to hear the other's suggestions before moving forward.

aashrav:
I think we can implement the variable limit later because I don't believe it is a relevant issue right now. Also i think it will be easier to implement after we get some frontend components set up.

api/main_endpoints/routes/DoorCode.js
```diff
@@ -0,0 +1,105 @@
...
+const {
+  OK,
+  BAD_REQUEST,
+  UNAUTHORIZED,
+  FORBIDDEN,
+  NOT_FOUND,
+} = require('../../util/constants').STATUS_CODES;
+
+router.get('/getDoorCodes', (req, res) => {
+  DoorCode.find().then(doorcodes => res.status(OK).send(doorcodes));
```
jerrylee17, line 17: Might want a `.catch()` in case there's an error with fetching doorcodes.

api/main_endpoints/routes/DoorCode.js
```diff
@@ -0,0 +1,105 @@
...
+});
+
+router.get('/getAvailableDoorCode', (req, res) => {
+  DoorCode.findOne({ usersAssigned: { $lt: 2 } }).then((doorcodes) =>
```
jerrylee17, line 21: Add a .catch() for if there are no doorcodes left.

test/DoorCode.js
```diff
@@ -0,0 +1,257 @@
...
+    token: 'invalid',
+  };
+  const CODE_WITHOUT_REQUIRED_FIELDS = {};
+  const EDITED_CODE = {
+    doorCode: '051-4451',
+    doorCodeValidUntil: new Date('12/25/20'),
+    usersAssigned: 1,
+  };
+  const CODE_WITH_BAD_FORMAT = {
+    doorCode: '05121-44514211',
```
evanugarte, line 80: consider doing one more with a letter and symbol in it too

PR #538, opened 2020-07-06
Andrew/presubmit

Ajowyang:
created presubmit.py file which allows developer to run all checks(api, lint, frontend, build)
below is screenshot of terminal output to developer

Screen Shot 2020-07-06 at 3 56 38 PM

### DIFF
--- a/presubmit.py
+++ b/presubmit.py
@@ -1,0 +1,27 @@
+import os
+from subprocess import check_output, PIPE
+
+def run_test(test_name, command):
+    termCommand = command + ' ' + test_name
+    testName = test_name
+    numDots = 37 - len(testName)
+    dots = ""
+    i = 0
+    while i < numDots:
+        dots+="."
+        i = i + 1
+    
+    dotsPassed = dots + "passed"
+    dotsFailed = dots + "failed"
+    try:
+        check_output(termCommand, shell=True, stderr=PIPE)
+        print(test_name + dotsPassed)
+    except:
+        print(test_name + dotsFailed)
+    
+print("\nRunning tests....\n")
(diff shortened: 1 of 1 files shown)


presubmit.py
```diff
@@ -0,0 +1,28 @@
+import os
+from subprocess import check_output, PIPE
+
+print("\nRunning tests....\n")
+
+try:
+    check_output("npm run api-test", shell=True, stderr=PIPE)
```
evanugarte, line 7: nice! however imagine if we had 10 steps for presubmit, this file would grow a significantly in size. what about if we had 100 steps? this file would be 400+ lines long 😮 

consider making functions so we can avoid being redundant:

```py
def run_test(test_name, command):
  """
  In here you should put the try/except as well as the check_output function.
  params:
  test_name - the name of the test e.g. lint, build, etc.
  command - the command to be run by check_output e.g. npm run frontend-test
  """
  pass

```

presubmit.py
```diff
@@ -0,0 +1,27 @@
...
+
+def run_test(test_name, command):
+    termCommand = command + ' ' + test_name
+    testName = test_name
+    numDots = 37 - len(testName)
+    dots = ""
+    i = 0
+    while i < numDots:
+        dots+="."
+        i = i + 1
```
aashrav, line 12: instead of doing a while loop for the numDots, you can multiply a string in python by a number. For example you can do:  
`
variable = 'hello' * 2            //this is equivalent to saying variable = 'hellohello'

`
Try to implement this rather than using a while loop

evanugarte:
Closing this as we need to reopen in the [dev repo](https://github.com/SCE-Development/dev)

PR #545, opened 2020-07-10
Production Script

ThaiQ:


ThaiQ:
@qsaml 

> new coder question: why go back to an older version of `pdf-lib`, `reactstrap`, and `@babel/register`? or maybe I'm reading it wrong

 Ya right, fixed 

> from first look, would we then have to start running `npm run build` instead? will running server and frontend separately still work as before?

```npm run build``` wont be necessary

Server and frontend are still separated 
```npm run frontend``` to run the frontend
```npm server``` to run the back end

PR #561, opened 2020-07-21
Blast email api

justin-zhu1018:
Blast email backend

### DIFF
--- a/api/google_api/routes/Mailer.js
+++ b/api/google_api/routes/Mailer.js
@@ -2,6 +2,7 @@ const express = require('express');
 const router = express.Router();
 const { SceGoogleApiHandler } = require('../util/SceGoogleApiHandler');
 const { verification } = require('../email_templates/verification');
+const { blastEmail } = require('../email_templates/blastEmail');
 const {
   OK,
   BAD_REQUEST
@@ -9,11 +10,12 @@ const {
 const { googleApiKeys } = require('../../config/config.json');
 const { USER } = googleApiKeys;
 
+// Routing post /sendVerificationEmail calls the sendEmail function
+// and sends the verification email with the verification email template
 router.post('/sendVerificationEmail', async (req, res) => {
   const scopes = ['https://mail.google.com/'];
   const pathToToken = __dirname + '/../../config/token.json';
-  const apiHandler = new SceGoogleApiHandler(
-    scopes, pathToToken);
+  const apiHandler = new SceGoogleApiHandler(scopes, pathToToken);
   const tokenJson = await apiHandler.checkIfTokenFileExists();
 
   if (tokenJson) {
@@ -21,22 +23,45 @@ router.post('/sendVerificationEmail', async (req, res) => {
       apiHandler.refreshToken();
     }
   } else {
-    apiHandler.getNewToken(false);
+    apiHandler.getNewToken();
   }
 
-  await verification(
-    USER, req.body.recipientEmail, req.body.recipientName
+  await verification(USER, req.body.recipientEmail, req.body.recipientName)
+    .then((template) => {
+      apiHandler
+        .sendEmail(template)
+        .then((_) => {
+          res.sendStatus(OK);
+        });
+    })
+    .catch((_) => {
+      res.sendStatus(BAD_REQUEST);
+    });
+});
+
+// Routing post /sendBlastEmail calls the sendEmail function
+// and sends the blast email with the blast email template
+router.post('/sendBlastEmail', async (req, res) => {
+  const scopes = ['https://mail.google.com/'];
+  const pathToToken = __dirname + '/../../config/token.json';
+  const apiHandler = new SceGoogleApiHandler(scopes, pathToToken);
+
+  await blastEmail(
+    USER,
+    req.body.emailList,
+    req.body.subject,
+    req.body.content
   )
     .then((template) => {
-      apiHandler.sendEmail(template)
-        .then(_ => {
+      apiHandler
+        .sendEmail(template)
+        .then((_) => {
           res.sendStatus(OK);
-        })
-        .catch(_ => {
+        }).catch((_) => {
           res.sendStatus(BAD_REQUEST);
         });
     })
-    .catch(_ => {
+    .catch((_) => {
       res.sendStatus(BAD_REQUEST);
     });
 });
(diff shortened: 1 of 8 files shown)


api/google_api/routes/Mailer.js
```diff
@@ -2,41 +2,70 @@ const express = require('express');
...
-  await verification(
-    USER, req.body.recipientEmail, req.body.recipientName
+  await verification(USER, req.body.recipientEmail, req.body.recipientName)
+    .then((template) => {
+      apiHandler
+        .sendEmail(template)
+        .then((_) => {
+          res.sendStatus(OK);
+        })
+        .catch((_) => {
```
evanugarte, line 36: i think u can do a single `.catch` on the outside

api/main_endpoints/routes/Auth.js
```diff
@@ -81,15 +81,15 @@ router.post('/login', function(req, res) {
...
-                function(error, result) {
-                  if (error) {
-                    return res.sendStatus(BAD_REQUEST);
-                  }
-
-                  if (result.nModified < 1) {
-                    return res.sendStatus(NOT_FOUND);
-                  }
-                }
+                // function(error, result) {
```
evanugarte, line 84: is there a reason why this is commented out?

justin-zhu1018: This function was causing the server to crash every other time I compiled something in the api folder. Lawrence helped me with this on Slack a few days ago.

api/google_api/routes/Mailer.js
justin-zhu1018: do you want me to remove the inner  `.catch` for the /sendBlastEmail router post as well?

api/google_api/routes/Mailer.js
```diff
@@ -2,41 +2,67 @@ const express = require('express');
...
     .then((template) => {
-      apiHandler.sendEmail(template)
-        .then(_ => {
+      apiHandler
+        .sendEmail(template)
+        .then((_) => {
           res.sendStatus(OK);
         })
-        .catch(_ => {
+        .catch((_) => {
```
evanugarte, line 61: you can get rid of this inner `.catch` as the outer one will catch all errors

api/main_endpoints/routes/Auth.js
evanugarte: got it, you can remove this code then

src/APIFunctions/User.js
```diff
@@ -151,3 +152,35 @@ export async function checkIfUserExists(email) {
...
+
+/**
+ * This function takes in a list of current users and returns a
+ * filtered user list that is determined by the filter id.
+ * @param {array} users array of all registered users
+ * @param {integer} filterID represents what to filter email by
+ * @returns {array} filtered array of users
+ */
+export function filterUsers(users, filterID) {
+  let filteredUsers = users.filter((user) => {
```
evanugarte, line 164: `array.filter` will automatically omit elements from an array by us returning a boolean instead of `user` or `''`, read more [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)

src/APIFunctions/User.js
```diff
@@ -151,3 +152,35 @@ export async function checkIfUserExists(email) {
...
+    const d = new Date();
+    if (filterID === userFilterType.VALID) {
+      if (
```
evanugarte, line 167: you can instead do
```js
if (filterID === userFilterType.VALID) {
  return ( user.accessLevel >= membershipState.MEMBER
        && user.membershipValidUntil > d.toISOString() );
}
```

src/APIFunctions/User.js
```diff
@@ -151,3 +152,35 @@ export async function checkIfUserExists(email) {
...
+        user.accessLevel >= membershipState.MEMBER
+        && user.membershipValidUntil > d.toISOString()
+      ) {
+        return user;
+      }
+    } else if (filterID === userFilterType.NON_VALID) {
```
evanugarte, line 173: you can instead do
```js
} else if (filterID === userFilterType.NON_VALID) {
  return ( user.accessLevel === membershipState.NON_MEMBER ||
        user.accessLevel === membershipState.PENDING );
}
```

src/APIFunctions/User.js
```diff
@@ -151,3 +152,35 @@ export async function checkIfUserExists(email) {
...
+      if (
+        user.accessLevel === membershipState.NON_MEMBER ||
+        user.accessLevel === membershipState.PENDING
+      ) {
+        return user;
+      }
+    } else {
+      return user;
+    }
+    return '';
```
evanugarte, line 183: you can return nothing and the `filter` function will omit that element if thats what you were trying to do
```suggestion
```

api/google_api/routes/Mailer.js
justin-zhu1018: Removing the inner `.catch` caused an error in running the api-test so we decided to keep it for now. If there are any solutions, we can figure it out later

src/APIFunctions/Mailer.js
```diff
@@ -14,9 +14,38 @@ export async function sendVerificationEmail(email, firstName) {
...
+ * @param {string} subject - The subject of the email
+ * @param {string} content - The contents of the email
+ * @returns {ApiResponse} Containing any error information related to the
+ * request
+ */
+export async function sendBlastEmail(emailList, subject, content) {
+  let status = new ApiResponse();
+  await axios
+    .post(MAILER_API_URL + '/api/Mailer/sendBlastEmail', {
+      recipientEmail: emailList,
```
evanugarte, line 41: if it possible you chould change the backend to accept
`req.body.emailList, req.body.subject, req.body.content`?

that way you can do 
```js
.post('/path/to/api', {
  emailList,
  subject,
  content
})
```

justin-zhu1018: Got it, lmk if you want me to push the changes

evanugarte: yes please

api/google_api/routes/Mailer.js
evanugarte: no problem

jerrylee17:
Rebase

PR #573, opened 2020-07-30
APIFunction changes for print log

1234momo:
Added 2 functions into APIFunction.js for adding a print request and getting all the print requests

### DIFF
--- a/src/APIFunctions/2DPrinting.js
+++ b/src/APIFunctions/2DPrinting.js
@@ -81,13 +81,13 @@ export async function logPrintRequest(data) {
 
 /**
  * Retrieves all print logs
- * @returns {}  Containing all print logs
+ * @returns {ApiResponse}  Containing all print logs
  */
 export async function getAllLogs() {
   let status = new ApiResponse();
   await axios.get(LOGGING_API_URL + '/api/PrintLog/getPrintLogs')
     .then(response => {
-      status.data = response.data;
+      status.responseData = response.data;
     })
     .catch(() => {
       status.error = true;
--- a/src/APIFunctions/2DPrinting.js
+++ b/src/APIFunctions/2DPrinting.js
@@ -1,7 +1,8 @@
 import axios from 'axios';
 import {
   PrintApiResponse,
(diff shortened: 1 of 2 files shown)


src/APIFunctions/2DPrinting.js
```diff
@@ -1,6 +1,8 @@
 import axios from 'axios';
-import { PrintApiResponse } from './ApiResponses';
-import { RPC_API_URL } from '../config/config.json';
+import {
+  PrintApiResponse,
+  ApiResponse } from './ApiResponses';
```
hsorathia, line 4: have ApiResponse on it's own line so it looks like
```js
import {
  PrintApiResponse,
  ApiResponse
} from './ApiResponses';
```

evanugarte:
```suggestion
import {
  PrintApiResponse,
  ApiResponse
} from './ApiResponses';
```

src/APIFunctions/2DPrinting.js
```diff
@@ -62,6 +65,37 @@ export async function printPage(data) {
...
+/**
+ * Log the print request
+ * @param {Object} data         Encoded file
+ * @returns {PrintApiResponse}  Containing information for if
+ *                              the page is printing
+ */
+export async function logPrintRequest(data) {
+  let status = new ApiResponse();
+  await axios.post(LOGGING_API_URL + '/api/PrintLog/addPrintLog', data)
+    .catch(() => {
```
aashrav, line 77: If the routing doesn't throw an error what will this function return? I don't see status.responseData being set anywhere

src/APIFunctions/2DPrinting.js
```diff
@@ -62,6 +65,37 @@ export async function printPage(data) {
...
+      status.error = true;
```
aashrav, line 78: If the routing doesn't throw an error what will this function return? I don't see status.responseData being set anywhere

jerrylee17: I don't think there is any responseData needed. It just returns an ApiResponse with no error

PR #574, opened 2020-07-30
Added s3 bucket backend

thibui0908:
- Added a routing for getting a signed url in order to upload objects to S3 bucket.
- Added API functions in order to utilize getSignedUrl routing and upload objects.
- Created a test case in order to ensure routing behaves properly.


src/APIFunctions/S3Bucket.js
```diff
@@ -0,0 +1,36 @@
+import axios from 'axios';
+
+export function uploadToS3(files) {
+  for (let i = 0; i < files.length; i++) {
+    let file = files[i];
+    let fileParts = files[i].name.split('.');
```
evanugarte, line 6: is it possible that your backend could just take the filename as a whole (`sce.png`) and then handle the logic for splitting it there?

src/APIFunctions/S3Bucket.js
```diff
@@ -0,0 +1,36 @@
...
+    let fileName = fileParts[0];
+    let fileType = fileParts[1];
+    axios
+      .post('http://localhost:8082/api/S3Bucket/getSignedUrl', {
```
evanugarte, line 10: create a new url const string in 
[`src/config/config.example.json`](https://github.com/SCE-Development/Core-v4/tree/dev/src/config/config.example.json) and use it like we do in [`src/APIFunctions/LedSign.js`](https://github.com/SCE-Development/Core-v4/blob/89a77d8ea0d8e5ab125fc26331fcdb4993352167/src/APIFunctions/LedSign.js#L54)

src/APIFunctions/S3Bucket.js
```diff
@@ -0,0 +1,36 @@
...
+        fileName: fileName,
```
evanugarte, line 11: as mentioned in the comment before, lets just send a single param over `fileName` and handle the split over there.
```suggestion
        fileName,
```

src/APIFunctions/S3Bucket.js
```diff
@@ -0,0 +1,36 @@
...
+        const returnData = response.data.url;
+        const signedRequest = returnData.signedRequest;
+
+        const options = {
+          headers: {
+            'Content-Type': fileType,
+          },
+        };
+        axios
+          .put(signedRequest, file, options)
```
evanugarte, line 24: is this to download the file?

aashrav: This is to upload the file to S3. Does something look off?

src/APIFunctions/S3Bucket.js
aashrav: She should just use the Mailer url constant right...not create a new one?

src/APIFunctions/S3Bucket.js
thibui0908: I put it there since I needed the fileType down below at the headers. Shoud I still change it?

src/APIFunctions/S3Bucket.js
evanugarte: some comments explaining what's going on would be great

api/cloud_api/server.js
```diff
@@ -3,7 +3,8 @@ const { SceHttpServer } = require('../util/SceHttpServer');
 function main() {
   const API_ENDPOINTS = [
```
evanugarte, line 4: you should be able to do 
```suggestion
  const API_ENDPOINTS =  __dirname + '/routes`;
```

and you wont need to keep adding to the array.

src/APIFunctions/S3Bucket.js
```diff
@@ -0,0 +1,42 @@
+import axios from 'axios';
+import { MAILER_API_URL } from '../config/config.example.json';
+
+export function uploadToS3(files) {
```
evanugarte, line 4: 1. add a comment section above this function like other files to describe param types instead
1. you should be returning an `A[iResponse` object like other functions in this directory, take a look at any file for examples.

src/APIFunctions/S3Bucket.js
```diff
@@ -0,0 +1,42 @@
...
+    let file = files[i];
+    let fileParts = req.body.file.split('.');
+    let fileName = fileParts[0];
+    let fileType = fileParts[1];
+
+    // Retrieving signed url which will allow us to make an upload request to S3
+    axios
+      .post(MAILER_API_URL + '/api/S3Bucket/getSignedUrl', {
+        fileName: fileName,
+        fileType: fileType
```
evanugarte, line 15: remember this in the future for javascript
```suggestion
        fileName,
        fileType
```

src/Pages/Home/Home.js
```diff
@@ -17,7 +17,6 @@ import Iframe from 'react-iframe';
         display="initial"
         position="relative"
         allowFullScreen/>
-
```
evanugarte, line 20: revert changes in this file

src/APIFunctions/S3Bucket.js
```diff
@@ -0,0 +1,42 @@
+import axios from 'axios';
+import { MAILER_API_URL } from '../config/config.example.json';
+
+export function uploadToS3(files) {
+  for (let i = 0; i < files.length; i++) {
```
evanugarte, line 5: instead of a for loop consider doing [`array.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) so like

```js
files.map(file => {
  // the "file" variable is equivalent to files[i]
})
```

evanugarte:
@jerrylee17 @IsitaB please review before we merge

PR #579, opened 2020-08-03
Refactor printing page into components

### LINKED ISSUE #565: Refactor 2D Printing UI
We can split the page into multiple components like
```
<2DPrinting />
├── <Header />
├── <FileUpload />
├── <StatusModal />
├── <PrintPageModal />
└── <ConfirmationModal />
```
1. `FileUpload`: Contains the filepond and `PrintInfo` inside of it, also has the `Continue` button. When the user uploads a file we populate a hook in `2DPrinting`. Pressing `Continue` will show the `PrintPageModal`
1. `StatusModal`: Shows after user sends a print
1. `PrintPageModal`: Shows the file preview where u can specify page ranges and copies.
1. `ConfirmationModal`: Will show when user wants to print a page from the `PrintPageModal`

jerrylee17:
Resolves #565 

Also added descriptions of all the state variables so people can read it


src/Pages/2DPrinting/2D-printing.css
```diff
@@ -34,6 +34,7 @@
   transform: translateZ(0px);
   animation: fade-in 1s ease-in-out;
   animation-fill-mode: forwards;
+  cursor: pointer;
```
evanugarte, line 37: nice i learned about the types from you
image

jerrylee17: The front end component still works - I tested it. Stubbed the printingRPC return values. Making PR tonight on SCE-RPC to remove destination parameter.

I need to fix Travis tho

qsaml: so this would be the type of pointer that shows up while in the printing view window?

src/Pages/2DPrinting/2DPrintPageModal.js
```diff
@@ -0,0 +1,112 @@
...
+              </FormGroup>
+              <FormGroup check>
+                <Label check>
+                  <Input {...props.ModalBody.pagesSelect} />
+                  <Input {...props.ModalBody.pagesRange} />
+                </Label>
+              </FormGroup>
+            </FormGroup>
+            <Label {...props.ModalBody.legend}>
+              Note: All prints are black ink only
```
qsaml, line 99: personally sounds better if written as "Printouts are black and white"

src/Pages/2DPrinting/2DPrinting.js
```diff
@@ -194,6 +191,129 @@ export default function Printing(props) {
     setStatusModal(true);
   }
 
+  const printPageModalProps = {
+    isOpen: previewModal,
+    toggle: () => {
+      setPreviewModal(!previewModal);
+    },
+    size: 'xl',
+    ModalHeader: {
```
evanugarte, line 200: will most of this stuff ever change? you could just keep it in the component instead of sending it over.

evanugarte:
also add test cases!!!

src/Pages/2DPrinting/2DPrinting.js
jerrylee17: It works and prints. The styling components might change if we are to enhance the UI of this page (which we should do), but the functionalities won’t (cases where state variables are passed). I like keeping these together just in case we need to link events/styling together in the future.

src/Pages/2DPrinting/2DPrintPageModal.js
```diff
@@ -0,0 +1,112 @@
...
+  const {
+    loadPreview,
+    displayPagesLeft,
+    previewLabels,
+    setSides
+  } = props;
+
+  return (
+    <Modal {...props}>
+      <ModalHeader {...props.ModalHeader} >
```
hsorathia, line 26: the extra space after the `}` bothers me sm idk why

PR #588, opened 2020-08-07
Redo LED Sign Page

nanarboursal:
Creates cards for every message in the LED Sign message queue rotation


src/Routing.js
```diff
@@ -123,6 +116,10 @@ export default function Routing({ appProps }) {
     { Component: EventList, path: '/events' },
     { Component: OfficerDB, path: '/officerDB' },
     { Component: Team, path: '/team' },
+    {
```
IsitaB, line 119: make it one line

PR #604, opened 2020-08-24
update packages

jerrylee17:


hsorathia:
I get this error
image
when trying to run the site through `npm start`

hsorathia:
Got this error when trying to print:
image

But everything else works

jerrylee17:
> Got this error when trying to print:
> image
> 
> But this looks like a separate issue so it should be fine

Yeah this is not a package issue

PR #606, opened 2020-08-26
Added Access Level Verification

### LINKED ISSUE #577: Add Authorization to Backend APIs
Right now we don't check for access level in our protected APIs. This is easy to add, see below

1. We need a new parameter for `checkIfTokenValid`, `accessLevel` [`api/main_endpoints/util/token-functions.js`](https://github.com/SCE-Development/Core-v4/blob/55ad626d991b3048a917b63abb68620a89911dad/api/main_endpoints/util/token-functions.js#L23)

The access level should default to `MEMBERSHIP_STATE.MEMBER` so we dont have to send it in all the time
```js
function checkIfTokenValid(request, accessLevel=MEMBERSHIP_STATE.MEMBER) {
  // basically check if the user's access level >= accessLevel parameter
}
```
2. Refactor all the instances of `checkIfTokenValid` to send in the access level **don't use numbers!** use `MEMBERSHIP_STATE.ADMIN` as a parameter instead of something like `2` or `3`.

Example decoded response:
decoded

aashrav:
**Description**
-Added additional  checks in edit routing to make sure that user cannot alter their access level
-Login routing checks to make sure that user's email has been verified
-Tweaked the login error display for the frontend

**How to Test**

_Edit Routing_

- [x] Make sure you can edit users on admin account
- [x] Test a regular member account 
Get JWT token (print it in index.js)
Use postman and try to change accessLevel (should get unauthorized)
Use postman and edit an account with different email (should get unauthorized)
Use postman and make a valid edit

- [x] Test a Officer account 
Get JWT token (print it in index.js)
Use postman and try to change accessLevel to admin (should get unauthorized)
Use admin side of website and make a valid edit

_Email Verification_

- [x] Create account and try to login immediately after (make sure you get an error message)
- [x] Verify email and try logging in again

Resolves #577

### DIFF
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -9,7 +9,8 @@ const axios = require('axios');
 const { registerUser } = require('../util/registerUser');
 const {
   checkIfTokenSent,
-  checkIfTokenValid
+  checkIfTokenValid,
+  decodeToken
 } = require('../util/token-functions');
 const {
   OK,
@@ -55,7 +56,7 @@ router.post('/checkIfUserExists', (req, res) => {
 router.post('/delete', (req, res) => {
   if (!checkIfTokenSent(req)) {
     return res.sendStatus(FORBIDDEN);
-  } else if (!checkIfTokenValid(req, membershipState.ADMIN)) {
+  } else if (!checkIfTokenValid(req, membershipState.OFFICER)) {
     return res.sendStatus(UNAUTHORIZED);
   }
 
@@ -129,7 +130,9 @@ router.post('/users', function(req, res) {
   }
   User.find()
     .sort({ joinDate: -1 })
-    .then(items => res.status(OK).send(items))
+    .then(items => {
+      res.status(OK).send(items);
+    })
     .catch(() => {
       res.status(BAD_REQUEST).send({ message: 'Bad Request.' });
     });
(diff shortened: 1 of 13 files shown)


IsitaB (approved):
Rebase

api/main_endpoints/routes/User.js
```diff
@@ -55,7 +56,7 @@ router.post('/checkIfUserExists', (req, res) => {
 router.post('/delete', (req, res) => {
   if (!checkIfTokenSent(req)) {
     return res.sendStatus(FORBIDDEN);
-  } else if (!checkIfTokenValid(req)) {
+  } else if (!checkIfTokenValid(req, membershipState.OFFICER)) {
```
seb1055, line 59: I lack context here but it seems deleting a user should be reserved for Admin role. If officers have this privilege this is fine

api/main_endpoints/routes/User.js
```diff
@@ -142,6 +145,31 @@ router.post('/edit', (req, res) => {
...
+    }
+    if(req.body.accessLevel && req.body.accessLevel !== decoded.accessLevel){
+      return res
+        .status(UNAUTHORIZED)
+        .send('Unauthorized to change access level');
+    }
+  }
+
+  if(decoded.accessLevel === membershipState.OFFICER){
+    if(req.body.accessLevel && req.body.accessLevel == membershipState.ADMIN){
```
seb1055, line 168: I am not so sure about this line, nothing from the request body should be trusted server side. This logic look to first check the JWT token for the Officer access level (this is good) then check the request body for Admin level. This could allow an Officer to masquerade as a Admin.

seb1055 (changes requested):
Overall this is a major improvement! 

I **did not** see a change to /api/Mailer/sendBlastEmail which previously allows any authenticated user to send a email using the club email. I would include that fix in this pull request. 

Additionally the way role verification per route is done is very error prone. This should work for a hot fix but I would strongly recommended moving the logic to express middle-ware. This will standardize the logic and prevent people either forgetting the if statements or implemented them incorrectly.

api/main_endpoints/routes/User.js
seb1055: @aashrav clarified my misunderstanding here, resolving this.

jerrylee17:
rebase 
image

ThaiQ:
Good tested.

ThaiQ (approved):

seb1055 (approved):

PR #633, opened 2020-09-14
Fixes RESTful calls in src/apiFunctions

ThaiQ:
1. Fixes RESTful calls in src/apiFunctions to adapt nginx load-balancer.
2. Added `SceRPCServer` to `nginx.conf` at port `8083`


IsitaB (approved):
🔥

PR #668, opened 2020-10-18
fix printLogs null error - printing Analytics

1234momo:


IsitaB (approved):
works for me :)

aashrav (approved):
Works!

PR #674, opened 2020-10-26
Refactored UI and fixed tests for Events Page

Annabel1054:
**Before:** 
image

**After:** 
image
image

### DIFF
--- a/src/Pages/EventManager/EventManager.js
+++ b/src/Pages/EventManager/EventManager.js
@@ -70,6 +70,7 @@ export default function EventManager(props) {
         <Button className='create-event' onClick={toggleNewEvent}>
           New Event
         </Button>
+        <div> <br></br> </div>
         {modal && (
           <EventManagerModal
             modal={modal}
--- a/src/Pages/Events/EventCard.js
+++ b/src/Pages/Events/EventCard.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import { Row } from 'reactstrap';
 import { getDateWithSlashes } from '../../APIFunctions/Event';
-import { mapPinSymbol, clockSymbol } from '../Overview/SVG';
 
 function EventCard(props) {
   const {
     title,
+    description,
     eventDate,
     eventLocation,
     startTime,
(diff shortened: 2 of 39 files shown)


aashrav (changes requested):
Great Work! Just a couple things I think should be fixed

1.  Events that have already happened(before current date) should not be shown on the Event Page
2. Right now the events are shown in descending order meaning that events that are next month would render above the events that are tomorrow. This should be the other way around so that people can see which events are coming up on the top itself. 
3. This is a small issue of overflowing description when a word is too long. Can't think of when someone would enter such a large word, but defiantly something to consider 
image
4. Another suggestion is making the title a different color so that it stands out more.

src/Pages/EventManager/EventManager.js
```diff
@@ -70,6 +70,7 @@ export default function EventManager(props) {
         <Button className='create-event' onClick={toggleNewEvent}>
           New Event
         </Button>
+        <div> <br></br> </div>
```
law-nguyen, line 73: what is this for?

Annabel1054: It creates a space between the "new event" button and "click on any event below to edit or delete it"
image

law-nguyen (changes requested):
This is a great start!
1. I would address the changes that aash proposes.
2. when you shrink the page down, maybe changing the card layout to have the event on the bottom and the picture on top of it would be better. Right now the scaling is a bit off :<

image


I would propose adding functionality to swap displays to something like this when the width of the screen gets too small 
image

Annabel1054:
@aashrav the events are now chronologically ordered and the events that have passed will no longer show up on the events page.
image
image

Annabel1054:
@law-nguyen fixed!!
image

src/Pages/Events/EventCard.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 import { Row } from 'reactstrap';
```
law-nguyen, line 2: delete this line! `Row` is never used

src/Pages/Events/EventList.js
```diff
@@ -30,6 +32,24 @@ function AnnouncementList() {
...
+        currDate.setDate(currDate.getDate() -1);
+        let validList = [];
+        eventList.forEach(item => {
+          let date = new Date(item.eventDate);
+          if (date >= currDate) {
+            validList.push(item);
+          }
+        }, setValidList(validList), setGetFiltered(false));
+      } catch (error) {
+        const alertText = 'There are no events to filter!';
```
law-nguyen, line 48: This isn't used either in the code, so it's alright to take it out for now.

law-nguyen (changes requested):
Almost done! There are a couple of warnings that pop up when running build that you gotta fix.

image

IsitaB (changes requested):
Screen Shot 2020-11-26 at 11 22 51 AM
fix this

qsaml:
functionality of creating and displaying events works, but running `npm run frontend-test` on my end returns a few errors

evanugarte:
this is a big pr

PR #684, opened 2020-11-01
updated package.json for dependabot prs

hsorathia:
#670 #677 #678 #680 #681 #685 #687 #688 

haven't done the axios and react one because those ones will probably break something and we might need to change how we do things


hsorathia:
it works ! ! !

IsitaB (approved):

aashrav (approved):
works

PR #727, opened 2020-12-20
Added Alumni Tag

thibui0908:
- Add an alumni role to the account in Overview Page
- Enable sending email exclusively to Alumni

How to test:
- Comment out sendBlastMail function
- Use the code block to test on the console:
    /*eslint-disable*/
    emailList.forEach(item => {console.log(item)})

### DIFF
--- a/src/APIFunctions/User.js
+++ b/src/APIFunctions/User.js
@@ -173,7 +173,9 @@ export async function checkIfUserExists(email) {
 export function filterUsers(users, filterID) {
   let filteredUsers = users.filter((user) => {
     if (filterID === userFilterType.VALID) {
-      return (user.accessLevel >= membershipState.ALUMNI);
+      return (user.accessLevel >= membershipState.MEMBER ||
+        user.accessLevel === membershipState.ALUMNI
+      );
     } else if (filterID === userFilterType.NON_VALID) {
       return (
         user.accessLevel === membershipState.NON_MEMBER ||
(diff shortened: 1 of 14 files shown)


src/APIFunctions/User.js
```diff
@@ -173,7 +173,7 @@ export async function checkIfUserExists(email) {
 export function filterUsers(users, filterID) {
   let filteredUsers = users.filter((user) => {
     if (filterID === userFilterType.VALID) {
-      return (user.accessLevel >= membershipState.MEMBER);
+      return (user.accessLevel >= membershipState.ALUMNI);
```
IsitaB, line 176: this implementation would be problematic if a member specific functionality uses this function

IsitaB (approved):

src/APIFunctions/User.js
IsitaB: accesslevel would be a problem because changing the numbers to role mapping would invalidate the accessLevels of existing users

IsitaB: TokenFunctions.js line 10 may need updating

IsitaB: an alumn after auth would only have access to their profile and signedoutroutes

IsitaB (approved):
🔥

aashrav (approved):
Works for me! I noticed that when you try to send an email to "Everyone" it doesn't send an email to Alumni. After looking at the code I am assuming that is how it is supposed to be designed. I would suggest to rename "Everyone" to something else because it may be misleading for first time users

IsitaB (approved):

PR #733, opened 2020-12-24
Sticky Navbar 

fmasood23:


src/Components/Navbar/UserNavbar.js
```diff
@@ -81,12 +82,14 @@ export default function UserNavBar(props) {
...
 
+
   useEffect(() => {
     // eslint-disable-next-line
   }, [])
 
   return (
-    <div className='user-nav'>
+
+    <div className={'user-nav'}>
```
IsitaB, line 92: get rid of the curly braces

IsitaB (changes requested):
The navbar overlaps each page's first component. otw it works great :)

why do we need `margin-top:50px` in all these files? is there a better way to handle this?

src/Components/Navbar/UserNavbar.js
```diff
@@ -81,12 +81,14 @@ export default function UserNavBar(props) {
     setMenuIsOpen(!menuIsOpen);
   };
 
-  useEffect(() => {
+
+  useEffect(() =>
```
IsitaB, line 85: why do we need this? get rid of it

IsitaB (changes requested):
get rid of the hardcoded margin-tops

make the navbar's position:sticky instead

IsitaB (approved):

jerrylee17 (approved):

IsitaB (changes requested):
delete the empty useEffect block and take package-lock out of the PR since you havent made any dependency changes

IsitaB (approved):

PR #738, opened 2020-12-30
Refactored Events Page (Branch #2) -- Fixed text overlapping issue.

Annabel1054:


.gitignore
```diff
@@ -89,4 +89,3 @@ sheets.json
 *.csr
 *.key
 *.cer
```
IsitaB, line 91: dont include this file in the PR

src/Pages/EventManager/EventManager.js
```diff
@@ -70,6 +70,7 @@ export default function EventManager(props) {
         <Button className='create-event' onClick={toggleNewEvent}>
           New Event
         </Button>
+        <div> <br></br> </div>
```
IsitaB, line 73: add padding to the button instead

src/Pages/EventManager/EventManagerModal.js
```diff
@@ -203,6 +203,7 @@ function EventManagerModal(props) {
             <Label>Event Description</Label>
             <Input
               type='textarea'
+              maxLength = {80}
```
IsitaB, line 206: okay. is the height of the card hardcoded bc this seems to be taking care of long descriptions?

src/Pages/Events/EventCard.js
```diff
@@ -25,19 +24,20 @@ function EventCard(props) {
...
-        {getDateWithSlashes(eventDate.slice(0, 10))} {startTime} -
-        {' '}
-        {endTime}
-      </Row>
-      <Row className='event-location'>
-        {mapPinSymbol()}
-        {' '}
-        {eventLocation}
-      </Row>
+      <div className='event-text-block'>
```
IsitaB, line 27: text block position is hard coded not next to each other

test/frontend/EventList.test.js
```diff
@@ -50,23 +50,21 @@ describe('<EventList />', () => {
       ' array',
     async () => {
       returnEventArray();
```
IsitaB, line 52: add in a test for your filtering functionality

IsitaB (changes requested):

IsitaB (changes requested):
get rid of the onclick modal for Events.js

test/frontend/EventList.test.js
```diff
@@ -50,23 +50,21 @@ describe('<EventList />', () => {
...
+      expect(wrapper.find(EventList)).to.have.lengthOf(
         RENDERED_EVENTS.responseData.length
       );
     }
   );
   it('Should render a title if no events are returned', async () => {
     returnEmptyArray();
-    const wrapper = await mount(<EventList />);
+    const wrapper = await mount(<EventList {...RENDERED_EVENTS} />);
+    const result = wrapper.find('.no-event-list');
```
IsitaB, line 63: this isn't testing whether it would render 'no events yet'. you can add this code block to the previous test and it would pass

src/Pages/Events/EventList.js
```diff
@@ -30,38 +19,49 @@ function AnnouncementList() {
...
   }, []);
 
-  function handleClick(clickedEvent) {
-    setEvent(clickedEvent);
-    toggle();
-  }
-
   return (
-    <div className='event-background'>
+    <React.Fragment>
```
ThaiQ, line 47: Do you need ReactFragment here?

src/Pages/EventManager/EventManager.js
```diff
@@ -64,7 +64,7 @@ export default function EventManager(props) {
   }
 
   return (
-    <div className='event-background'>
+    <React.Fragment>
```
ThaiQ, line 67: Do you need React Fragment ?

ThaiQ (changes requested):
Dont commit package-lock.json, i will cause conflicts with others dev

ThaiQ:
ignore the React.Fragment comments

IsitaB (changes requested):
the edit modal is dead for EventManager. I cant edit/delete events anymore

IsitaB (approved):
great work :)

ThaiQ (approved):

aashrav (changes requested):
When you try to create an event it doesn't let you create it until the required fields are filled in. But if you type into a required field and then delete it, the create event button is still enabled. When you create the event it throws an error, but this could be designed better. Instead I believe that the button should always be enabled and if you try to create an event without required fields a message should show up on the modal saying "Required fields aren't filled in".

IsitaB (changes requested):
yooo look at those form validation errors :')

functionality wise: you dont get the same form validation errors when youre trying to submit
Screen Shot 2021-01-05 at 11 46 25 PM
in this screenshot, my title field is empty and it doesnt let me submit, but also doesnt tell me whats wrong

UI wise: the comment could look better; the dash dash isnt great
Screen Shot 2021-01-05 at 11 47 05 PM
`<fieldname> is required` or `please input <fieldname>`should suffice

IsitaB (commented):
it works!!

but the modal code's repetitive 
this would be better https://serverless-stack.com/chapters/create-a-custom-react-hook-to-handle-form-fields.html

law-nguyen (approved):
it works for me

fmasood23 (approved):
it works for me as well!! looks good!

aashrav (approved):
Looks good. One thing I think should be fixed is that the events on the event manager should show up in chronological order like how they do on the actual events page.

charlotte-zhuang (commented):
I was having issues with input validation for the time and date (I'm not sure if that's a separate issue). Overall, it looks alright to me.

On a side note, are the delete and cancel buttons are backwards in the ConfirmationModel component?

## Comments

1. Entering `12:00` as a time turns into 12:00 am. Also, `24:00` turns into 12:00 pm (but that's not a valid time anyway).
2. Invalid times are not rejected, e.g. `10:00 foo` or `hello`
3. Input validation errors in general are not mentioned (unless you leave a field blank).
4. Missing placeholder text for time and date.
5. I think using index for a component's key might cause issues (I think that's the default behavior if you don't specify a key)? I thought I ran into an error while deleting an event, but I couldn't reproduce it.
6. Creating or editing an event takes two clicks (this might be a problem with my setup)

ThaiQ (approved):

PR #836, opened 2021-04-27
Dependabots

IsitaB:


IsitaB:
tests pass. works for me

charlotte-zhuang (approved):
Tests passed and pages look okay. Events page was buggy, but I think it was always buggy.

Could we take this opportunity to commit `api/package-lock.json` and fix the `@tinymce/tinymce-react` version in `package.json`?

charlotte-zhuang (approved):
I did a quick check after merging with `dev` and it looks ok.

PR #894, opened 2021-08-07
Count time

### LINKED ISSUE #892: Create method to collect user time traffic
To collect user time traffic, this method will need to run on all webpages regardless of the permission level. It will also need to send a JSON Object in the specified format below:
```javascript
{
   "PageName" : string,
   "StartTime" : string,
   "EndTime" : string,
   "UserID" : string,
   "SSOID" : string
}
```

neelp03:
This pull-request adds features to collect user traffic data, send collected data to aws lambda shredder and resolve #892. These new features collect data which will allow SCE to visualize website traffic. 
Possible future features include adding middleware which will act as a bridge between applications.


andrewtedi:
Closed for now, refer back to other PRs from the same time using the summer21internship tag

PR #897, opened 2021-09-04
Undo SCE microservice backend

evanugarte:
The SCE backend now runs on a single port. This is so all services can be reached by the user's browser.

**TL;DR:** we have 1 backend server instead of 4.

### Visual helpers
Moves the architecture from this:
old-arch
to this:
SCE Website Architecture drawio
The `api/` folder now has the below structure, merging all the `routes`, `models` and `util` directories.
image

### Why does this pr exist?
TODO create explanation @evanugarte


api/util/logging-helpers.js
```diff
@@ -9,16 +9,24 @@ const axios = require('axios');
...
- * @returns {boolean} - Whether the save was successul or not
+ * @returns {Promise<boolean>} - Whether the save was successul or not
  */
 async function addErrorLog(errorToAdd) {
-  let errorSaved = true;
-  await axios.post(LOGGING_API_URL + '/api/ErrorLog/addErrorLog',
-    { ...errorToAdd })
-    .catch(err => {
-      errorSaved = false;
+  const newError = new ErrorLog({
```
evanugarte, line 15: this functions in this file are updated to save stuff to mongodb directly, instead of making an http request

evanugarte:
we may be able to use the nginx conf to fix the microservices instead. closing for now

PR #900, opened 2021-10-27
Use REST API instead of gRPC for LED Sign

andrewtedi:
hotdog

Uses Express and Axios HTTP requests to connect to LED Sign server. Passes both LED sign health check and update when tested.


evanugarte:
i believe we can close this? due to #908

PR #901, opened 2021-11-19
ESlint errors don't stop website from running anymore

allen-tran:
See this [stackoverflow answer](https://stackoverflow.com/a/64630234). `react-scripts` newer version causes eslint to run automatically with `npm run start`, causing the website to crash if there are any lint errors.

This PR prevents that from happening. Lint errors will be caught by PR builds and `npm run lint` instead.

See below example for lint failing:

Screen Shot 2021-11-18 at 5 53 15 PM


evanugarte (approved):
Verified this in person

andrewtedi (approved):

PR #906, opened 2021-12-03
Rewrite emails for vice president and president to main SCE email

andrewtedi:
uh yeah
fortnite-default


allen-tran:
rollie gang

PR #912, opened 2021-12-07
Fixes eslint package file from preventing start

allen-tran:
This PR fixes a dependency issue with our package.json. Trying to run Core-v4 was resulting in this error:

image


evanugarte (approved):
Looks good. We tested this earlier and it worked

andrewtedi (approved):
Wow! Amazing! Great work Allen!

PR #916, opened 2021-12-13
Optimizes Frontend Docker Build

### LINKED ISSUE #914: Optimizes Frontend Docker Builder
In [`docker/Dockerfile-frontend:13`](https://github.com/SCE-Development/Core-v4/blob/dev/docker/Dockerfile-frontend#L13) we are copying the entire contents of Core-v4 into our frontend container. The only directories we really need are `public`, `util` and `src`. Changing the Dockerfile to follow this will decrease the container size and shorten build times.

imdominicreed:
Resolves #914

### DIFF
--- a/docker/Dockerfile-frontend
+++ b/docker/Dockerfile-frontend
@@ -4,7 +4,7 @@ RUN mkdir -p /usr/src/app
 
 WORKDIR /usr/src/app
 COPY package.json ./
-RUN npm install
+RUN npm install --production
 COPY public ./public
 COPY util ./util
 COPY src ./src
--- a/docker/Dockerfile-frontend
+++ b/docker/Dockerfile-frontend
@@ -2,15 +2,13 @@ FROM node:lts-alpine
 
 RUN mkdir -p /usr/src/app
 
-COPY src /usr/src/app/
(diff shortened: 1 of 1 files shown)


evanugarte (approved):
### Original buildtime:
image

### New buildtime:
image


Over a minute saved! Furthermore if the image is built once then rebuilding it will also save time. For example, here is if I change a file and rebuild:

image

Only 90 seconds! This is a great start, many more revisions to come for sure.

docker/Dockerfile-frontend
```diff
@@ -2,15 +2,13 @@ FROM node:lts-alpine
...
 
-COPY src /usr/src/app/
-COPY public /usr/src/app/
-COPY util /usr/src/app/
-COPY package.json /usr/src/app/
-
 WORKDIR /usr/src/app
-
+COPY package.json ./
 RUN npm install
```
evanugarte, line 7:
```suggestion
RUN npm install --production
```

allen-tran (approved):
oh my gosh this code is so hot 😩💦

allen-tran:
are u single???

imdominicreed:
for you i am 😳

PR #918, opened 2021-12-21
mongodb connection issue resolved

ajha7:
MongoDB issue with using localhost for dbHost, replacing it with 127.0.0.1 seems to work.


evanugarte:
lint is failing, check build output for the fix

allen-tran (changes requested):
as mentioned previously,linting

evanugarte (approved):

allen-tran (approved):
looks so good to me! wonderful job u are like a SRE

PR #936, opened 2021-12-28
Adds checks for Google API keys

neelp03:
```js
const { APIKeys} = require('config.json');
const { SECRET, ID } = APIKeys;

if(SECRET != 'NOT_SET' && ID != 'NOT_SET') {
  executeFile();
} else {
   doNothing();
   //if its a promise then resolve
}
```


api/cloud_api/routes/Mailer.js
```diff
@@ -10,6 +10,9 @@ const {
 const { googleApiKeys } = require('../../config/config.json');
 const { USER } = googleApiKeys;
 
+/**
+ * Check if API keys are valid
+ */
```
evanugarte, line 15:
```suggestion
```

api/cloud_api/util/SceGoogleApiHandler.js
```diff
@@ -41,16 +42,19 @@ class SceGoogleApiHandler {
...
+    }
   }
 
   /**
-   * Checks if the token file from the given path from the constructor exists.
-   * It will either resolve a boolean false or an object containing the token
-   * data.
-   * @returns {Promise<boolean|Object>} The result of the token file search.
-   */
+     * Checks if the token file from the given path from the constructor exists.
```
evanugarte, line 53: this file is full of random changes, i would turn ur autoformatter off and undo these

api/cloud_api/util/SceGoogleApiHandler.js
```diff
@@ -41,16 +42,19 @@ class SceGoogleApiHandler {
     );
     this.oAuth2Client.setCredentials({
       // eslint-disable-next-line
-      refresh_token: REFRESH_TOKEN
+        refresh_token: REFRESH_TOKEN
     });
+    if(CLIENT_ID != 'NOT_SET' && CLIENT_SECRET != 'NOT_SET') {
```
evanugarte, line 47: what if we only say they have valid api keys if theyre both set?
```suggestion
    if(CLIENT_ID != 'NOT_SET' || CLIENT_SECRET != 'NOT_SET') {
```

src/APIFunctions/Mailer.js
```diff
@@ -79,7 +79,7 @@ export async function addEventToCalendar(newEvent, token) {
     baseURL: MAILER_API_URL
   });
   await firstAPI
-    .post('api/Calendar/addEventToCalendar', { token, newEvent })
+    .post('/Calendar/addEventToCalendar', { token, newEvent })
```
evanugarte, line 82: good job catching this

evanugarte (commented):
A good start, we discussed always resolving Promises because the client will otherwise be left hanging.

Need some handlers in the calendar/gmail function to resolve true if the api keys are not valid

api/cloud_api/util/SceGoogleApiHandler.js
```diff
@@ -141,23 +144,27 @@ class SceGoogleApiHandler {
...
-      }, (err, res) => {
-        if (err) return reject(false);
-        const events = res.data.items;
-        if (events.length) {
-          resolve(events);
-        } else {
-          reject(false);
-        }
-      });
+      if(this.hasValidAPIKeys) {
```
evanugarte, line 147: instead of wrapping the functions in an if/else, at the top of each you can do

```js
if (!this.hasValidAPIKeys) {
  return resolve(true);
}
// rest of original function
```

evanugarte (approved):
image

allen-tran (approved):
🔛🔝🔜

PR #933, opened 2021-12-26
Shows upcoming events on public events page

neelp03:
### Resolves #926 
- created API that retrieves all upcoming events (including today's)
- created API function for `events/getUpcomingEvents`
- updated public event page to populate the page with all upcoming events by calling `getUpcomingEvents()` api function

> Note: Public view shows only upcoming events while admin view (event manager) shows all events as seen below (look at the dates).

PR #938, opened 2021-12-30
Removes filtered events function

allen-tran:
With @neelp03 integrating a fresh, [new way of organizing the events list in a way that makes sense](https://github.com/SCE-Development/Core-v4/pull/933), we can remove the unnecessary function `getFilteredEvents` along with its corresponding states.

### DIFF
--- a/src/Pages/Events/EventList.js
+++ b/src/Pages/Events/EventList.js
@@ -34,13 +34,13 @@ function AnnouncementList() {
             return (
               <EventCard
                 key={index}
-                isEventManager={false}
+                isEventManager = {false}
                 {...event}
               />
             );
           })
         ) : (
-          <h1 className='no-event-list'>No events yet!</h1>
+          <h1 className = 'no-event-list'>No events yet!</h1>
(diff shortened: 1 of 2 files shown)


evanugarte (commented):
what did getFilteredEvents even do?

src/Pages/Events/EventList.js
```diff
@@ -19,23 +18,6 @@ function AnnouncementList() {
...
-          let date = new Date(item.eventDate);
-          if (date >= currDate) {
-            validList.push(item);
-          }
-        }, setValidList(validList), setGetFiltered(false));
-      } catch (error) {
-      }
-    }
-  };
-
```
neelp03, line 38: deleting this causes an error in the frontend-test for some reason

allen-tran: i am changing the test for eventlist and refactoring it a bit

evanugarte (approved):

neelp03 (approved):

PR #940, opened 2021-12-31
Removes running in production checks

neelp03:
The `api\cloud_api\util\SceGoogleApiHandler.js` file was checking if it was running in production, which disabled sending emails for development. Pull Request: https://github.com/SCE-Development/Core-v4/pull/936 adds a better condition for when not to be able to work with mailer and calendar.


evanugarte (approved):

imdominicreed (approved):
lgtm

Issue #922, opened 2021-12-22
Add Logic for API Key not Being Present


ajha7:
# Add Logic for API Key not Being Present

A lot of our code assumes the API key to be valid at runtime.
[`api/cloud_api/routes/Mailer.js`](https://github.com/SCE-Development/Core-v4/blob/66fbb80a6180f8aaf38a2b162a4aaa3f04924616/api/cloud_api/routes/Mailer.js)
is an example, where it imports and uses the keys. We should add logic for checking if
the keys are valid and if they aren't, disable the service.
 
### How to do this
- [x] Change all default config.example.json from `XXXXX...` to some default value like `NOT_SET`
- [x] Add logic for checking if a config.example.json value is equal to `NOT_SET`, and if it isn't, don't run the service.

neelp03:
Added api checks for Google API keys, Discord API keys, Github API keys, and s3Bucket Access keys.

PR #945, opened 2022-01-01
Adds checks for Github API Keys

neelp03:
### Resolves part of #922 
- checks is client ID and client secret are changed from the default value before using them
- changes default values in `config.example.json` to "NOT_SET"
```js
const {
  CLIENT_ID,
  CLIENT_SECRET
} = require('config.json').githubApiKeys;

if(CLIENT_ID == 'NOT_SET' && CLIENT_SECRET == 'NOT_SET') {
  stopRunningThisFunction();
} else {
   doWhatever();
}
```
> Note: current default value: 'XXXX...X'    --->    New default value: 'NOT_SET' in config.example.json


evanugarte (approved):
Very clean change,!! In the future core-v4 should log something like

```
skipping <insert thing it was trying to do> because the API key was not set.
```

this can be done by someone else, let's find something more fun for you to work on!

allen-tran (approved):
awesome

PR #946, opened 2022-01-01
Fix Create New Event Double Click

allen-tran:
## Having to click twice for creating a new event is nevermore.
- previously, `Create New Event` button in EventManager.js requires you to click twice. @evanugarte noticed that props are async and thus the logic for creating sequentially was broken. 
- this bug was introduced by @evanugarte with #225. very sad.

**BEFORE**: This button was broken. See Below

image

**AFTER**: This button is not broken. See Below

image

## Features
- [X] reduces required clicks from 2 to 1!!!! 💯 🔛 🔝 🔜


evanugarte (approved):
> Allen! You did it Allen! You did what I [could not](https://www.youtube.com/watch?v=VqW_LX8_zFw&t=89s)!

neelp03 (approved):
LGTM!
That button was so annoying. Thanks for fixing it.

PR #908, opened 2021-12-04
Using SQS to send writing requests to LED Sign

andrewtedi:
Since Core-v4 cannot directly talk to the VM for our LED Sign, SQS can be used to link the two servers. Our Core-v4 backend would make requests to our LED Sign SQS queue and the VM would read any data sent to said queue and use the data in updating the LED Sign.

PR #949, opened 2022-01-01
Add const declaration to LED AWS creds 

neelp03:
If no var, let, or const statement is used then the variable gets assigned to the global scope or reassigned where it is found in the scope, which very often is not the desired behavior, since that leads to cluttering in the scope, or variable overriding.

Previously: 
```js
creds = new AWS.Credentials(ledSqsKeys.CLIENT_ID, ledSqsKeys.CLIENT_SECRET);
```
Now:
```js
const creds = new AWS.Credentials(ledSqsKeys.CLIENT_ID, ledSqsKeys.CLIENT_SECRET);
```


neelp03:
basically had to create a new branch and open this pr just because @allen-tran didn't fix it in #908  😑

evanugarte (approved):
allen is on thin ice

allen-tran:
> basically had to create a new branch and open this pr just because @allen-tran didn't fix it in #908  😑 

🆗

allen-tran (approved):

PR #953, opened 2022-01-10
Created schema for reset token

crowbardispensingcore:


api/main_endpoints/models/password-reset.js
```diff
@@ -0,0 +1,23 @@
+const mongoose = require('mongoose');
+const Schema = mongoose.Schema;
+
+const PasswordReset = new Schema(
+  {
+    email: {
```
allen-tran, line 6: two things:
- [ ] the definition of the schema should follow the naming conventions of the other schemas
- [ ] the file name should also follow the naming convention of the other files in the models folder

api/main_endpoints/models/password-reset.js
```diff
@@ -0,0 +1,23 @@
+const mongoose = require('mongoose');
```
allen-tran, line 1: this only change should be this single file, you can undo the other changes by resetting it with dev with `git checkout origin/dev filename`

api/main_endpoints/models/password-reset.js
```diff
@@ -0,0 +1,23 @@
...
+      required: true
+    },
+    createdAt: {
+      type: Date,
+      default: Date.now,
+      required: true,
+      index: {  expires: '1h' }
+    }
+  }
+);
```
allen-tran, line 21: this is solid tho and ily

allen-tran (changes requested):

package.json
```diff
@@ -72,6 +72,7 @@
     "react-scripts": "^4.0.3",
     "reactstrap": "^8.10.0",
     "serve": "^11.3.2",
+    "server": "^1.0.37",
```
neelp03, line 75: package.json should remain the same as dev right?

neelp03 (changes requested):

PR #963, opened 2022-01-14
Add backend for password reset

crowbardispensingcore:
Based on PR #962 
New endpoints:

- `/request-reset`
  ```json
  {
    "email": "example@email.com"
  }
  ```
  The endpoint is called when the user submit their email in `/forgot-password` page. If the requested email exists, a reset token will be generated and inserted to `passwordresets` #962. An email containing the reset link will be sent to the user. The record is removed after 1 hour by MongoDB. The status code is always `OK` if no error occurred.

- `/validate-reset-token`
  ```json
  {
    "token": "exampletoken"
  }
  ```
  This endpoint is called when the user open the reset link sent to their email.

- `/reset-password`
  ```json
  {
    "newPassword": "123456",
    "email": "example@email.com",
    "token": "exampletoken"
  }
  ```
  This endpoint is called when the user submit their new password `/password-reset` page. Password will only be updated if the new password satisfies the server-side validation and email-token pair matches the record in `passwordresets` collection. The corresponding document will be removed if the password is successfully updated.


api/main_endpoints/util/password.js
```diff
@@ -0,0 +1,52 @@
+const bcrypt = require('bcrypt');
+
+function validatePassword(password, strength = 'medium') {
```
evanugarte, line 3: password strength functions already exist in [`api/main_endpoints/util/registerUser.js`](https://github.com/SCE-Development/Core-v4/blob/89a77d8ea0d8e5ab125fc26331fcdb4993352167/api/main_endpoints/util/registerUser.js)

you can add the hash password in there

api/main_endpoints/routes/Auth.js
```diff
@@ -238,4 +240,147 @@ router.post('/validateVerificationEmail', async (req, res) => {
...
+    return null;
+  }
+}
+
+function generateResetLink(token) {
+  const link = 'http://localhost:3000/reset-password/?token=' + token;
+  return link;
+}
+
+const { sendResetEmail } = require('../../cloud_api/util/mailer');
```
evanugarte, line 270: you can't import functions from a different api service, as they are seperate entities in prod.

instead u can make a call to the cloud api like [`api/main_endpoints/util/logging-helpers.js`](https://github.com/SCE-Development/Core-v4/blob/b1634aeb0f8808361e75821001aa72531e6d072b/api/main_endpoints/util/logging-helpers.js)

with peripheral apis. lmk if this makes sense

PR #973, opened 2022-01-27
2D and 3D printing pages

Megan-J:
updated the designs of the 2D and 3D printing pages

3d printing:
image

2d printing:
image

### How to test
- run frontend, backend and mongodb
- sign in as a user with at least member priviledge
- navigate to the 2d and 3d printing pages with expected visuals above


src/Pages/3DPrinting/3DPrintForm.js
```diff
@@ -9,11 +9,18 @@ import {
...
+  Col,
 } from 'reactstrap';
 import { submit3DPrintRequest } from '../../APIFunctions/3DPrinting';
-import Header from
-'../../Components/Header/Header.js';
+import mailImage from '../3DPrinting/gmail-icon.png';
+import discordImage from '../3DPrinting/discord-icon.png';
+
+export const copyrightMessage = `Created with ❤️ by the SCE Development 
+Team | © 2021 Software and Computer Engineering Society at SJSU`;
```
evanugarte, line 19: we dont have a copyright lol

to make this follow the current year no matter what, do

```suggestion
Team | Software and Computer Engineering Society at SJSU
 ${new Date().getFullYear()}`;
```

src/Pages/3DPrinting/3D-print-form.css
```diff
@@ -1,13 +1,20 @@
-.App {
+@import url('https://fonts.googleapis.com/css2?family=IBM+Plex+Mono:wght@300&family=Mina:wght@400;700&family=Ubuntu:wght@700&display=swap');
```
evanugarte, line 1: Is it possible we can use a different font that doesn't require a url? here are a bunch of css built-in fonts ([w3.org](https://www.w3.org/Style/Examples/007/fonts.en.html))

andrewtedi (commented):
Looks good! Is it possible that instead of an input for Additional Comments that a textarea can be used instead? The box itself is too long and borders of the box reach over the container. Also, experiment with adding breaks after the copyright message, since there isn't enough extra space at the bottom.

evanugarte (approved):
Verified this locally, great work!!

jessabelledelen (changes requested):
Hi Megan :) Good start and great job implementing the design! Just a couple things:

Was wondering why we're opting for drop-down lists for color & reason for usage instad of the original radio buttons design? 🥺 also wondering what this would mean for when a color is out of stock. Also, was wondering why we're doing a URL for .stl file versus a file upload.
They both work! Just wondering why we're swapping.

Category highlight is a little confusing -- it's okay that we're doing different tab designs from the Figma, but if the Current page we're on could be the tab with the lighter color to signify that it's selected, would be less confusing / more intuitive

Category tabs, and page contents like the text fields / dropdown lists shouldn't span entire width of the page / it'd be nice if we could focus them towards the center like the Figma design 

Footer should be at the absolute bottom of the page rather than the bottom of the page contents. With that, we could improve the vertical spacing of the elements and make it more uniform?

Again, wonderful start, Megan :) ❤️ Looking forward to future updates!!

Love, Jessabelle!!!

Megan-J:
@jessabelledelen 

Hi! Earlier in the semester, I showed my designs to Allen and Andrew and we agreed that the drop-downs would be easier to implement currently. If you still want to do the buttons, I can change the code. I chose to go with the URL for the .stl file versus a file upload since the backend currently isn't set up to support file uploads. 

I'll fix the other issues! :)

jessabelledelen (approved):
owowowo great job!!! ❤️ hard work pays off ily!!

.eslintrc.json
```diff
@@ -19,7 +19,6 @@
         "requireConfigFile": "false",
         "sourceType": "module"
     },
-    "parser": "@babel/eslint-parser",
```
andrewtedi, line 22: pls add it back tyty

andrewtedi (approved):

PR #975, opened 2022-01-27
Create RFID schema

neelp03:
```js
const RFIDSchema = new Schema(
  {
    name: {
      type: String,
      required: true,
    },
    byte: {
      type: String,
      unique: true,
      required: true,
    },
    created: {
      type: Date,
      default: Date.now,
    }
  },
  { collection: 'RFID' }
);
```


api/peripheral_api/models/RFID.js
```diff
@@ -0,0 +1,27 @@
+const mongoose = require('mongoose');
+const Schema = mongoose.Schema;
+
+const RFIDSchema = new Schema(
+  {
+    name: {
+      type: String,
+      required: true,
+    },
+    byte: {
```
evanugarte, line 10: maybe give a more descriptive thing like `cardData`? or add a comment above this line explaining what `byte` is
```suggestion
    // the actual bytes of a user's RFID card
    byte: {
```

api/peripheral_api/models/RFID.js
```diff
@@ -0,0 +1,27 @@
...
+    },
+    lastScanned : {
+      type: Date,
+      default: Date.now,
+    }
+  },
+  { collection: 'RFID' }
+);
+
+module.exports = mongoose.model('RFID', RFIDSchema);
```
evanugarte:
```suggestion
module.exports = mongoose.model('RFID', RfidSchema);

```

evanugarte, line 27:
```suggestion
module.exports = mongoose.model('RFID', RfidSchema);
```

api/peripheral_api/models/RFID.js
```diff
@@ -0,0 +1,27 @@
...
+      default: Date.now,
+    },
+    lastScanned : {
+      type: Date,
+      default: Date.now,
+    }
+  },
+  { collection: 'RFID' }
+);
+
```
evanugarte, line 26: It would be best if we don't store the user's true card data, but instead a hash of it.

This can look something like:

```js
RfidSchema.pre('save', function(next) {
  if (this.isModified('byte') || this.isNew) {
    bcrypt.genSalt(10, function(error, salt) {
      if (error) {
        return next(error);
      }
      bcrypt.hash(this.byte, salt, function(error, hash) {
        if (error) {
          return next(error);
        }
        this.byte = hash;
      });
    });
  }
  return next();
});
```

also would need custom logic to compare a card with the hash. I stole the above example from [`api/main_endpoints/models/User.js`](https://github.com/SCE-Development/Core-v4/blob/371f90a139ad30132155c2ab7e0eaf2ac9926376/api/main_endpoints/models/User.js#L77). You should also grab the compare logic from there too.

neelp03: So when the user scans the card we will have to hash the incoming byte and then look for the hashed byte in the database

evanugarte: yes sir, similar to how we do passwords in the above link

evanugarte: heres an example of us comparing passwords when the user logs in: [`api/main_endpoints/routes/Auth.js#L83`](https://github.com/SCE-Development/Core-v4/blob/371f90a139ad30132155c2ab7e0eaf2ac9926376/api/main_endpoints/routes/Auth.js#L83)

haaydenjones: Storing the information w/ a hash sounds like it would be optimal for what we need. +1

neelp03: OKay I will implement it before making pr for the API handlers and pub-sub

evanugarte (approved):
lgtm

haaydenjones (approved):
Looks good
lgtm

PR #976, opened 2022-01-28
generate-token script saves refresh_token

crowbardispensingcore:
Fixed a bug where the refresh token isn't saved properly in `config.json`

### How to test
- [ ] setup google api tokens following https://github.com/SCE-Development/Core-v4/wiki/Getting-Started#setting-up-mailer 
- [ ] run `npm run generate-token` from Core-v4 repo
- [ ] enter respective tokens following prompts
- [ ] finish the script
- [ ] inspect `token.json` and `config.json` to verify if the respective fields have been updated with the entered values
values to inspect:
```json
{
    "access_token": "access_token",
    "scope": "scope",
    "token_type": "token_type",
    "expiry_date":100000,
    "refresh_token": "refresh_token"
}
```
```json
{
  "googleApiKeys": {
    "CLIENT_ID": "what you entered",
    "CLIENT_SECRET": "what you entered",
    "REDIRECT_URIS": [
      "https://developers.google.com/oauthplayground"
    ],
    "REFRESH_TOKEN": "same as token.json",
    "USER": "idk"
  }
}
```


neelp03 (changes requested):
The link that is given in the command line after providing the client id and client secret, does not update the client id parameter to the given client id

That link looks something like this: 

> `https://accounts.google.com/o/oauth2/v2/..../client_id=THIS_NEEDS_TO_CHANGE&redirect_uri=https%3A%2F%2Fdevelopers.google.com%2Foauthplayground`

Another thing I noticed was that the refresh token property in config.json does not get updated. 
> Note: After running the script, a token.js file is generated which has the refresh token.

api/cloud_api/util/AuthManager.js
```diff
@@ -55,19 +55,71 @@ class AuthManager {
...
+          `Current Client ID: ${yellowColor}` +
+          `${config.googleApiKeys.CLIENT_ID}${defaultColor}`
+        );
+        while ((input.toLocaleLowerCase() !== 'y') &&
+               (input.toLocaleLowerCase() !== 'n')) {
+          input = await this.inputPrompt(
+            'Do you want to enter a new Client ID (Y/N)? '
+          );
+        }
+      }
```
evanugarte, line 75: image

neelp03 (approved):

andrewtedi (approved):

evanugarte (approved):
Tested this locally with big paul and it works

PR #977, opened 2022-01-28
Give Eslint warnings instead of errors when running website

andrewtedi:
Currently, if you run the frontend of the website with linting errors, you will receive an error screen preventing you from interacting with the website. We do not want this to happen. The fix that we have proposed is for Eslint to send warnings rather than errors. This will reserve linting errors solely for merging purposes and we do not need to fix linting errors to run the website, which could be useful for debugging.

How to test:

In order to fully test this PR, the use of three or more different OS's (Windows, MacOS, Linux) is required. 

Steps for Windows:
1. Run `python3 setup.py` in the Core-v4 directory
2. In Windows Powershell, run `Get-ChildItem Env:` This will display all the environment variables in your system.
3. If you find ESLINT_NO_DEV_ERRORS and it is set to true, then the setup ran successfully!

Steps for Linux and Mac:
1. Run `python3 setup.py` in the Core-v4 directory
2. Run `echo $ESLINT_NO_DEV_ERRORS` in a new terminal, which should return true.

Notes:
- To further check if the changes work, make a console.log in the code of any page or make any other linting error that would be caught after running the steps. If the website loads and your terminal gives Eslint warnings rather than errors, the setup was successful.
- Eslint warnings are yellow in the terminal while errors are red.
- For multiple runs of setup.py for Linux and Mac users, you may need to edit certain files to remove excess duplicate code. For Linux users, the file that you will need to edit to remove said code is either .bashrc or .bash_profile. For Mac users, the file is either .bashrc or .zshrc.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,38 @@
+import subprocess
 import platform
 import os
+
+def write_alias_to_file(file_name):
+    try:
+        subprocess.check_call(
+            f'grep -rl \"alias ESLINT_NO_DEV_ERRORS=true\"',
+            stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+    except subprocess.CalledProcessError:
+        with open(file_name, 'a') as file:
+            file.write('\n')
+            file.write(f"export ESLINT_NO_DEV_ERRORS=true")
+            file.write('\n')
+
+def add_alias_unix():
+    HOME_PATH = os.environ["HOME"]
+    BASHRC_PATH = f'{HOME_PATH}/.bashrc'
+    if platform.system() == "Darwin":
+        ZSHRC_PATH = f"{HOME_PATH}/.zshrc"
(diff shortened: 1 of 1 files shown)


ajha7 (approved):

setup.py
```diff
@@ -1,6 +1,33 @@
+import subprocess
 import platform
 import os
 
+def write_to_file(file_name):
```
evanugarte, line 5: seems like this function will run every time setup is ran. is it possible we can bring back the logic to do this conditionally? something like

```py
def alias_already_exists(path_to_config)
  with open(path_to_config, 'r') as f:
      for line in f.readlines():
        if 'export ESLINT_NO_DEV_ERRORS=true' in line:
          return True
  return False
```

andrewtedi: The last two commits should now resolve this.

evanugarte (approved):

Issue #1015, opened 2022-03-07
Add token verification to LED sign API


evanugarte:
basically #442 but in [`api/peripheral_api/routes/LedSign.js`](https://github.com/SCE-Development/Core-v4/blob/370ba6774cb7ebe8a732d455304b807609ef91bf/api/peripheral_api/routes/LedSign.js). we don't check tokens atm meaning anyone who knows `curl` or Postman can write whatever they want to our LED sign.

c00lme:
are we able to use a JSON web token for this?

c00lme:
and verify

neelp03:
Yes, that's what we are using for verification. The logic behind JWT verification for the peripheral server was implemented in PR #1027 
Refer to this: https://github.com/SCE-Development/Core-v4/commit/c04343c5da063113e13ebb34ef52c2a48a3f81da

Issue #1004, opened 2022-02-25
Make print request handling a single endpoint


evanugarte:
When the user prints a page, much of the logic to log the print request and update the user's print is handled in the front end. This means that a savvy user can figure out how to print unlimited pages since the printing API doesn't check the user's available page count at the moment.

We should make printing a page become a single API call to `/sendPrintRequest`, defined in [`api/peripheral_api/routes/Printer.js#L24`](https://github.com/SCE-Development/Core-v4/blob/291138b5291685399e0b393b7b9a03fe8d7bfeed/api/peripheral_api/routes/Printer.js#L24). The API should handle logging the request, subtracting the user's page count and sending the print request to SQS.

### How to do this
- We should first abstract much of the API's code to helper functions. For example logging the print request, pushing data to SQS and subtracting from the user's print count.
- The API should make use of a token to allow only true members to be able to print.
- The API should first check if the request is able to be satisfied given how many pages a user has left within a week. If not, we return a `400` with error information.

It is imperative that we do this ASAP since its a hole in only allowing members to print.

PR #1027, opened 2022-03-11
Make token-verification a REST endpoint

### LINKED ISSUE #1026: Making token verification a REST endpoint
RFID server uses token functions to verify user membership before creating or deleting an RFID. Token-function uses passport.js which is located in `main_endpoints/util`. https://github.com/SCE-Development/Core-v4/blob/d0eb4fadc91da5f939c2b03613d81107cc24d161/api/peripheral_api/routes/RFID.js#L7 https://github.com/SCE-Development/Core-v4/blob/d0eb4fadc91da5f939c2b03613d81107cc24d161/api/util/token-functions.js#L6
Problem is that, when in production, `peripheral_api/*` has no way to access `main_endpoints/*` which will break production.  Making token verification an endpoint will allow usage of token verification in peripheral server

Use the endpoint in Auth.js for token verification and modify API tests to adapt to the new changes.

neelp03:
Resolves #1026


api/util/passport.js
```diff
@@ -1,7 +1,7 @@
 const JwtStrategy = require('passport-jwt').Strategy;
 const ExtractJwt = require('passport-jwt').ExtractJwt;
-const User = require('../models/User');
-const { secretKey } = require('../../config/config.json');
+const User = require('../main_endpoints/models/User');
```
evanugarte, line 3: this is inside main_endpoints, meaning we will still get the below error in docker:
image

we have two options:
1. move the user schema to util/
2. make token verification a REST endpoint (`/api/verifyToken`) for other services outside of `main_endpoints` to call on.

1 would be much easier than 2. however the schema would be in an uncommon place. your call.

evanugarte (approved):
Tested locally and it works. This is very useful for #1004 and #1015

api/util/passport.js
neelp03: option 2 is the move since it will help later with other token verification in peripheral server

andrewtedi (approved):
Great work Neel!

PR #1031, opened 2022-03-14
fixing verification issues

Megan-J:
this fixes the member verification issue where people could not log in after making an account and verifying

### How to test
- [x] create a pending user with `npm run create-user`
- [x] ensure the pending user cannot sign in
- [x] create a non-member user with `npm run create-user` again
- [x] ensure the non member user can sign in and does **not** have access to any member services


evanugarte:
What does this pr do? My vision might be bad, I don't see a description

andrewtedi (approved):
All tasks accomplished on my end! Great job Megan!!!

evanugarte (approved):
Motivational. Consider opening an issue to better handle pending/non verified account sign in, right now we don't explain why they can't sign in.

PR #1069, opened 2022-04-05
Adds dynamic URL to peripheral token verification

### LINKED ISSUE #1068: Conditionally change token verification endpoint URL
Prod build peripheral APIs that use token verification break because of this one line.

# How to fix:

Add a condition that checks whether it is in prod or not which the post url will depend on

https://github.com/SCE-Development/Core-v4/blob/fd8eb91428428390741998d953d42c08dfb08650/api/util/token-verification.js#L12

Something like this:

```js
const mainEndpointUrl = MAIN_ENDPOINT_URL || MAIN_ENDPOINT_PROD
// can just be mainendpoints:8080 if its prod else localhost:8080
   // ...
   axios
  .post(`${mainEndpointUrl}/api/Auth/verify`, { token })
```

neelp03:
Resolves #1068

### DIFF
--- a/api/util/token-verification.js
+++ b/api/util/token-verification.js
@@ -1,6 +1,9 @@
 const axios = require('axios');
 const membershipState = require('./constants').MEMBERSHIP_STATE;
-let MAIN_ENDPOINT_URL = process.env.NODE_ENV === 'production' ?
+// If we are in a docker/prod environment, we can't rely on localhost
+// to route the request to the Auth API so we use the name of
+// the container instead,
+const MAIN_ENDPOINT_URL = process.env.NODE_ENV === 'production' ?
   'mainendpoints:8080' : 'localhost:8080';
 
 /**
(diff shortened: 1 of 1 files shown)


api/util/token-verification.js
```diff
@@ -1,5 +1,7 @@
 const axios = require('axios');
 const membershipState = require('./constants').MEMBERSHIP_STATE;
+let MAIN_ENDPOINT_URL = process.env.NODE_ENV === 'production' ?
+  'mainendpoints:8080' : 'localhost:8080';
```
evanugarte, line 4:
```suggestion
// If we are in a docker/prod environment, we can't rely on localhost
// to route the request to the Auth API so we use the name of
// the container instead,
const MAIN_ENDPOINT_URL = process.env.NODE_ENV === 'production' ?
  'mainendpoints:8080' : 'localhost:8080';
```

evanugarte (approved):

WillN-byte (approved):

PR #1086, opened 2022-04-21
Remove unnecessary images 

neelp03:
We don't need it since the frontend running in the Docker container


evanugarte (approved):
We fixed this by using a _____ to ____ the SJSU network from ______

CCLDArjun (commented):
lgtm

CCLDArjun (approved):
oops I had meant to approve

PR #1088, opened 2022-04-24
Fixed nginx issues

neelp03:
For some reason, all APIs stopped working. These changes should fix it..

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -6,18 +6,17 @@ http {
     #logging
     access_log  /var/log/nginx/sce.access.log;
 
-    #upstream connection to node servers from npm run server
     upstream webserver {
-        server frontend:3000;
+        server frontend.core-v4_default:3000;
     }
     upstream mainendpoints {
-        server mainendpoints:8080;
+        server mainendpoints.core-v4_default:8080;
     }
     upstream sce-cloud-api {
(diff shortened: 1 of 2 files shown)


nginx.conf
```diff
@@ -53,10 +52,10 @@ http {
...
         }
-	location / {
+        location / {
             proxy_pass http://webserver;
         }
     }
 }
 
-events { }
+events { }
```
evanugarte, line 61:
```suggestion
events { }

```

nginx.conf
```diff
@@ -6,18 +6,17 @@ http {
     #logging
     access_log  /var/log/nginx/sce.access.log;
 
-    #upstream connection to node servers from npm run server
```
evanugarte, line 9: paste this somewhere in both `.conf` files for reference:
```nginx
# when we run docker-compose up for the first time, a default
# network is created that all the containers are attached to.
# this default network name is the directory name + "_default".
# As a result, the network name will be "core-v4_default".
# To establish a connection to the containers, we can do
# <container name>.<network name>, and that's what happens below.
```

evanugarte (approved):
Thank you @aaj-gd for the assist

imdominicreed (approved):
BING CHILLING 🥶

PR #1092, opened 2022-05-30
New UI for Membership Application page

KuroyukiRyuu:
Redesigned membership application page while also removing the previous pages showing membership plans


neelp03 (changes requested):
🔥 This looks so good. Don't forget to resolve merge conflicts. I do have a few suggestions. I think the membership plans visual could be a bit smaller for mobile view but its up to you.

There is also a lot of commented out code (mostly in the css files) which should be removed.
You should also stick to the naming conventions we have been using. JS files would follow pascal case which is like this: "PascalCase" and CSS files are camel case which is like this: "camelCase".

neelp03:
One more thing, don't commit/push package.json and package-lock.json unless you are using new dependencies.

neelp03 (changes requested):
Membership app tests lgtm! I think you missed one file while renaming to follow the appropriate naming convention ([src/Pages/MembershipApplication/membershipApplication.jsx](https://github.com/SCE-Development/Core-v4/blob/022c46343959e76063d3f71db33b011e685e1904/src/Pages/MembershipApplication/membershipApplication.jsx)). It should follow pascal case which is basically camel case but it starts with uppercase.

src/Pages/MembershipApplication/registerPage.css
```diff
@@ -0,0 +1,277 @@
+@import url('https://fonts.googleapis.com/css2?family=Mina&display=swap%27');
+
+body {
```
evanugarte, line 3: in react, adding a rule for `body` in one css file will apply it to the whole project, same as the rules below for h1, h2, etc.

to fix this lets not use generic selectors and instead use class names, a good example of this is [src/Pages/3DPrinting/3D-print-form.css:27](https://github.com/SCE-Development/Core-v4/blob/053a0dae3380f93319107e9ebd43e61d1fab07b2/src/Pages/3DPrinting/3D-print-form.css#L27)

neelp03 (changes requested):
Every once in a while api-tests in github actions or checks will fail. If it seems like an odd error like something that is completely unrelated to your changes then just re-run all jobs via checks tab which should fix it. The frontend tests failed because you still have a path to a deleted file in your member app tests so removing that line should fix it.

neelp03 (approved):

evanugarte:
This looks good, 2 things
1. the name inputs arent aligned like the others
2. white area at the bottom depending on window size

i can show u how to fix these
image

evanugarte (approved):
summer walker once said
image

neelp03 (approved):

PR #1094, opened 2022-06-04
Fix printer pages reset function

ShaivPandya:
Reset was not functioning properly so the Federal Reserve fixed it.
Delete duplicate function in Users.js.


neelp03 (approved):

evanugarte (approved):
tested and it works, good stuff

PR #1111, opened 2022-07-05
Update AdminDashboard UI

davidtanline:
Screen Shot 2022-07-04 at 3 59 50 PM
Screen Shot 2022-07-04 at 4 06 18 PM

### DIFF
--- a/src/Components/Navbar/AdminNavbar.js
+++ b/src/Components/Navbar/AdminNavbar.js
@@ -14,7 +14,7 @@ export default function AdminNavbar(props) {
   const [collapsed, setCollapsed] = useState(true);
   const navbarLinks = [
     { title: 'Home', route: '/' },
-    { title: 'Overview', route: '/dashboard' },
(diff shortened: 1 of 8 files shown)


neelp03 (changes requested):
I think you should change the background to a dark color so the theme is consistent with all of our other pages but overall looks good.

neelp03 (changes requested):
Nice, it looks better with the dark bg color. You should also make the dashboard items smaller (see attached image). You also have some unused imports which have to be removed. When I click the dashboard items, it doesn't actually send me to the corresponding page, what it did was change the url only. The url changes but its not actually rendering that page (or loading that page, equivalent to entering a url and clicking enter). Lmk if that makes sense or not.

image

davidtanline:
Screen Shot 2022-07-06 at 8 01 02 PM

Updated UI, made items smaller

ShaivPandya (approved):

neelp03 (approved):
I'm with it.

PR #1126, opened 2022-07-22
Fix Homepage Bug

neelp03:
There is a bug on the homepage that messed up the whole frontend for mobile devices
Ideally when sliding horizontally, CSS will take care of the logic and will not show the whitespace underneath
But the homepage image frame was setting a margin which pushed it off of the screen causing it to show the white space when sliding horizontally. I fixed it by adding padding instead of a margin.

p.s. most was just trial and error 💀


src/Components/Frame/HeroFrame.css
```diff
@@ -54,21 +54,36 @@
...
+    }
+    .hero-image {
+        margin-top: 0px !important;
+        border-radius: 10px;
+    }
+    .hero-frame {
+        padding: 10px;
+        width: 369px;
+    }
+}
```
evanugarte, line 89:
```suggestion
}

```

evanugarte (approved):
certified neel classic

upload a screenshot of what stuff looks like now

ShaivPandya (approved):

PR #1134, opened 2022-08-04
Door code discord bot

fahmed0710:
Core-v4 API to retrieve doorcodes based on user's discord ID


neelp03 (commented):
There a bunch of random changes, most likely because of auto formatter
fyi, if you are using prettier with vscode in the settings.json of vscode, you can add this line:
```json
"prettier.singleQuote": true
```
so when you do auto format it keeps single quotes, similarly you can do
```json
  "trailingComma" : "none"
```
to avoid adding random commas when you do auto format, p sure there are also ways to avoid adding space before function parentheses when doing auto format

neelp03 (commented):
No API tests?

neelp03 (changes requested):
Merge conflicts and consider adding api tests

ShaivPandya (approved):

PR #1140, opened 2022-08-20
Add helper functions for updating LED sign state through SSH tunnel

akshtshrma24:


api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,20 @@
+async function updateSignAxios(axios, req) {
+    await axios
+        .post('http://host.docker.internal:11000/api/update-sign' + {...req.body})
+        .then(() => {
+            return true
+        }).catch((err) => {
+            return false;
+        });
+}
+async function healthCheckAxios(axios) {
```
evanugarte, line 10: why are you passing axios as a param? just import it at the library at the top
```suggestion
async function healthCheck() {
```

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,20 @@
+async function updateSignAxios(axios, req) {
```
evanugarte, line 1: send in ...req.body as a param to updateSign

```suggestion
async function updateSign(data) {
```

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,20 @@
...
+            return true
+        }).catch((err) => {
+            return false;
+        });
+}
+async function healthCheckAxios(axios) {
+    await axios
+        .get('http://host.docker.internal:11000/api/health-check')
+        .then(() => {
+            return true
```
evanugarte, line 14: this wont work

api/peripheral_api/routes/LedSign.js
```diff
@@ -12,8 +11,7 @@ const {
   checkIfTokenSent
 } = require('../../util/token-verification');
 const logger = require('../../util/logger');
-
-const SqsHandler = new SceSqsApiHandler(AWS.Queue.LED_QUEUE_NAME);
+const axiosFunc = require('../util/LedSign')
```
evanugarte, line 14:
```suggestion
const { updateSign, healthCheck } = require('../util/LedSign')
```

evanugarte (approved):

api/peripheral_api/routes/LedSign.js
```diff
@@ -23,14 +21,7 @@ router.get('/healthCheck', async (req, res) => {
...
   }
-  await axios
-    .get('http://host.docker.internal:11000/api/health-check')
-    .then(() => {
-      return res.sendStatus(OK);
-    })
-    .catch((err) => {
-      return res.sendStatus(NOT_FOUND);
-    });
+  await updateSign(...req.body);
```
evanugarte, line 24: switch these, we are calling update sign in health check and health check in update sign

api/peripheral_api/routes/LedSign.js
```diff
@@ -47,12 +38,10 @@ router.post('/updateSignText', async (req, res) => {
...
   }
-  const result = await SqsHandler.pushMessageToQueue(req.body);
-  if (result) {
-    res.sendStatus(OK);
-  } else {
-    res.sendStatus(BAD_REQUEST);
+  if (process.env.NODE_ENV !== 'production') {
+    return res.sendStatus(OK);
   }
+  await healthCheck();
```
evanugarte, line 44: you also need to check the return value of these, if we are up return 200 else 500

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,34 @@
+const axios = require('axios').default;
+
+async function updateSign(req) {
```
evanugarte, line 3: replace all the `return`s with `resolve`s

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,34 @@
+const axios = require('axios').default;
```
evanugarte, line 1:
```suggestion
const axios = require('axios');
```

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,34 @@
+const axios = require('axios');
+
+async function updateSign(req) {
```
evanugarte, line 3:
```suggestion
async function updateSign(data) {
```

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,34 @@
...
+  return new Promise((resolve) => {
+    try {
+      axios
+        .post('http://host.docker.internal:11000/api/update-sign' + {req})
```
evanugarte, line 7:
```suggestion
        .post('http://host.docker.internal:11000/api/update-sign' + {...data})
```

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,34 @@
+const axios = require('axios');
+
+async function updateSign(req) {
+  return new Promise((resolve) => {
+    try {
```
evanugarte, line 5: the try catch is unnecessary here because youre already doing `.catch()` on the axios call. remove for both

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,34 @@
+const axios = require('axios');
```
evanugarte, line 1: import logging and log when the `catch` happens like
```
logging.error("healthcheck/updateSign had error:", e);
```

allen-tran (approved):
this might be troll but every file in this directory has JSDoc comments like 

```
 /**
   * Helper function to perform a health check on fades.
   * @returns {bool} If we can give a taper,
   * else false for cleaning clippers.
   */
```

so maybe we can add that

everyday we lit

api/peripheral_api/util/LedSign.js
```diff
@@ -0,0 +1,84 @@
...
+const logger = require('../../util/logger');
+
+/**
+ * These function is meant only for use in production, where the
```
evanugarte, line 5:
```suggestion
 * These functions are meant only for use in production, where the
```

PR #1163, opened 2022-09-19
Remove plugins section from .babelrc

akshtshrma24:
Removes noise from frontend tests as seen below:

image


davidtanline (approved):
slayyy

PR #1172, opened 2022-10-13
replace getUserByEmail with getUserById

danknessdra:
removed searchUserByEmail and replaced its uses with getUserByID for security

searchUserByEmail can be abused with different email reqs


akshtshrma24 (approved):
bobo

neelp03 (approved):
some random changes probably by auto formatter, you can change it back to whatever it was before if you want but looks good!

ShaivPandya (approved):

PR #1174, opened 2022-10-18
Remove Profile Image

akshtshrma24:
Remove Profile Image from nav bar
image

Before it was 
image

Profile Image had no use, no one else can see them, and no way to set them. Removed.


neelp03 (approved):

danknessdra (approved):
OK

PR #1198, opened 2022-12-18
Member Page for SCE members and non-member page/ message for non members

danknessdra:
This was done by changing the endpoint to give information to users with member membership status and above as opposed to officer. 

There is also a message for non members who want to access the member page.
nonmember:
image
member:
image

### DIFF
--- a/src/Pages/Profile/MemberView/Profile.js
+++ b/src/Pages/Profile/MemberView/Profile.js
@@ -114,6 +114,14 @@ export default class Profile extends Component {
       ]
       : [
         { title: 'Become a member to access this page!', value: '' },
+        { title: '', value: '' },
+        { title: '', value: '' },
+        { title: '', value: '' },
+        { title: '', value: '' },
+        { title: '', value: '' },
+        { title: '', value: '' },
+        { title: '', value: '' },
+        { title: '', value: '' }
       ];
 
     return (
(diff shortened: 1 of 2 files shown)


src/Pages/Profile/MemberView/Profile.js
```diff
@@ -110,18 +110,18 @@ export default class Profile extends Component {
...
         { title: '', value: '' },
-        { title: 'Door Code', value: '' },
-        { title: 'Joined Date', value: '' },
-        { title: 'Email', value: '' },
-        { title: 'Membership Expiration', value: '' },
-        { title: '2D Prints', value: '' },
-        { title: 'Request 3D Printing', value: '' },
-        { title: 'Sign-in with Discord', value: '' },
-        { title: 'Change Password', value: '' }
+        { title: '', value: '' }
```
evanugarte, line 124: instead of empty title and value can this array just be 1 element? or does it not work

evanugarte (approved):

src/Pages/Profile/MemberView/Profile.js
danknessdra: it fails the test if it is not 1 element

evanugarte: maybe change the test instead? up to u i already approved

danknessdra: sick bet

PR #1238, opened 2023-03-13
Npm run start windows support

Luceium:
Running npm run start has issues on windows if using the most recent Node.js.
This push is an alternative too downgrading Node.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "api-test": "mocha test/api/*.js",
     "frontend-test": "mocha --require @babel/register --require ignore-styles test/frontend/*.js --exit",
-    "start": "react-scripts start",
+    "start": "react-scripts --openssl-legacy-provider start",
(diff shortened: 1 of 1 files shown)


Luceium:
twas a bad solution. It allows npm start to run independent of node version however sce run c no longer worked (which is arguably more important)

Luceium:
@akshtshrma24 what do you think of creating a new script in pacakge.json
`"start-frontend": "react-scripts --openssl-legacy-provider start"`
Or alternatively, should we add in the clark wiki that users should download nvm. We could add the following steps to the wiki

1. after installing the latest version of Node go to [nvm's github page](www.github.com/coreybutler/nvm-windows)
2. Then click on the latest release 
image
3. If you are a windows user download the following asset 
image
4. run the setup wizard
5. run the following commands in a power shell terminal (note: you must open the terminal AFTER the setup wizard
```
nvm install v16.0.0
nvm use 16.0.0
```

PR #1242, opened 2023-03-22
Os dependent hot reload

Luceium:
Creates an environment variable folder based on your terminal's operating system when you run setup.py. The environment variables modify the behavior of hot reload. Without the two environment variables running a Linux-based docker image on a Windows device causes issues because the default Linux file watching is not compatible with the file system of the device. Because the files are hosted on the device and not in the Linux-based docker image, windows users need to override the default file watching necessary for hot reload.
If you are using a Mac or Linux-based device the Python script will set the environment variables to false (the default) causing nothing to change.

PR #1246, opened 2023-04-09
Background

Luceium:
Adds \<background/\> element that can be placed with no side effects.
The background has multiple optional props to modify it as desired.
All props have defaults and are a multiplier of the size. Notable defaults will be listed.
- size: changes the size of the background _(defaults to min(screen height, screen width)_
- innerRadius: controls size of inner square
- outerRadius: controls size of outer square
- radius: controls how round corners are _(defaults to 0)_
- lineSpread: controls seperation between circuit lines
- endDistance: controls how far out the circuit lines go
- lineStartOffset: allows circuit lines to start past the inner square

image


Luceium:
Note that I used #1242 to develop so #1242 needs to be approved first

danknessdra (approved):
LGTM

PR #1253, opened 2023-04-27
navbar sign in buttons

najm-masri:
shifted the sign in and sign up down to be centered with the other buttons.


src/Components/Navbar/navbar.css
```diff
@@ -50,6 +52,7 @@
 
 .sce-nav {
   flex-flow:row wrap;
+  padding-right: 121s0px;
```
jerrylee17, line 55: 121px?

jerrylee17 (changes requested):

PR #1283, opened 2023-09-24
Return the Sign up Page

akshtshrma24:
image

Sign up page, needs to be placed back. We still want people to sign up even if we are rewriting the website, we sitll need our old website to have full functionality.

image


pnavab (approved):
looks good but add the previous message back
let members know they have to pay and add my discord

jangomax (approved):

src/Pages/MembershipApplication/MembershipForm.js
```diff
@@ -285,7 +285,7 @@ export default function MembershipForm(props) {
           </div>
           <div className='venmo-link'>
             <a
-              href='https://venmo.com/u/sce-treasurer'
+              href='https://venmo.com/u/pnavab'
```
Luceium, line 288: restore this too sce-treasurer bc we pablo changed his venmo username

Luceium (commented):
lgtm. Let's test sign up works before pushing. Users were reporting issues with sign up.

PR #1301, opened 2023-10-15
Use environment variables for Peripheral API URLs

evanugarte:
every service behind an ssh tunnel now has the url specified as an environment variable

SCE Website Architecture drawio


danknessdra (approved):
yes good amazing LGTM

pnavab (approved):
good change

PR #1310, opened 2023-11-06
Finish Pagination for User Page

evanugarte:
1 year and some change late @davidtanline , 

running here https://sce.sjsu.edu/user-manager

looks like this:

image


joshua-demo (approved):
lgtm

PR #1353, opened 2024-01-11
Printing Page UI Height Fix

joshua-demo:
just changed the height of the printing screen so that its not the full height, but height minus nav bar height


evanugarte (commented):
the height looks kind of short for viewing a pdf? here is the current branch
image

here is dev:
image

evanugarte (approved):
looks good, please rebase your branch and add a screenshot of the new ui

banyar-shin (approved):
looks good to me!

PR #1365, opened 2024-01-20
add searching to cleezy

gursheyss:
image

### DIFF
--- a/src/APIFunctions/Cleezy.js
+++ b/src/APIFunctions/Cleezy.js
@@ -6,13 +6,13 @@ let PERIPHERAL_API_URL = process.env.REACT_APP_PERIPHERAL_API_URL
 
 export async function getAllUrls(token, page, search) {
   let status = new ApiResponse();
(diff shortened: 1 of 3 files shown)


src/APIFunctions/Cleezy.js
```diff
@@ -4,11 +4,15 @@ import { ApiResponse } from './ApiResponses';
...
+export async function getAllUrls(token, page, search) {
   let status = new ApiResponse();
+  let params = { token, page };
+  if (search !== undefined) {
+    params.search = search;
+  }
   await axios
     .get(PERIPHERAL_API_URL + '/Cleezy/list', {
-      params: { token, page }
+      params: params
```
evanugarte, line 15: i think we can do the same inline stuff we did in my other comment

evanugarte (commented):
add a screenshot of what the ui looks like

evanugarte (approved):
if this was tested locally and it works looks good

PR #1369, opened 2024-01-26
Clean up Cleezy page

pnavab:
- Add restrictions to Cleezy search to only allow alphanumeric search queries
- Move error alert message to the top of page, not only inside conditionally rendered div
- Simplify maybeRenderErrorAlert method to render an error message from a hook, the hook is set every time a boolean hook is on error mode

image


evanugarte (approved):
clean

akshtshrma24 (approved):

PR #1373, opened 2024-02-01
add internship info in about page

liubrian267:


evanugarte:
will resolve this with #1417

Issue #1375, opened 2024-02-01
Searching on Cleezy page bugged


pnavab:
URL shortener page should not clear the query on enter

Page should also default to a null search and list all when a blank search is searched

PR #1376, opened 2024-02-04
Fixed cleezy page searching

syng828:
Changed search query to null by default, lists all if empty, and removed clearing of query.
Resolves issue #1375

### DIFF
--- a/src/Pages/URLShortener/URLShortener.js
+++ b/src/Pages/URLShortener/URLShortener.js
@@ -19,7 +19,7 @@ export default function URLShortenerPage(props) {
   const [page, setPage] = useState(0);
   const [total, setTotal] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(10);
-  const [searchQuery, setSearchQuery] = useState('');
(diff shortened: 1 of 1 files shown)


src/Pages/URLShortener/URLShortener.js
```diff
@@ -85,7 +85,10 @@ export default function URLShortenerPage(props) {
   async function maybeSubmitSearch() {
     console.debug('reached this call');
     const regex = /^[a-zA-Z0-9]+$/;
-    if (regex.test(searchQuery)) {
+    if (searchQuery === '') {
+      setInvalidSearch(false);
+      getCleezyUrls(undefined, searchQuery);
```
pnavab, line 90:
```suggestion
      getCleezyUrls(page, searchQuery);
```
it should go back to the page they were on instead of undefined

syng828: fixed

pnavab (approved):
looks good

gursheyss (approved):
awsome

PR #1377, opened 2024-02-05
fix member profile page not loading

banyar-shin:
- make a new route for fetching user data
- fix expiration date for non-officers not being returned


joshua-demo (approved):
lgtm

PR #1378, opened 2024-02-05
fix profile not showing for pending & non-members

banyar-shin:
- fix profile page not loading for non-members and pending members


joshua-demo (approved):
so much work in this pr

lgtm

pnavab (approved):
why was this not in the first pr? i guess it is ok

PR #1382, opened 2024-02-21
2 21 2024 home page horizontal scroll bar fix

joshua-demo:
Removing the horizontal scroll bar when animating the home page


pnavab (approved):
why are we using .css files in 2024

steeevin88 (approved):
nice

Issue #1389, opened 2024-03-28
Add sorting for users


syng828:
Implement sorting for column in users table 
- By date joined
- By email
- By name

evanugarte:
closed by #1402

PR #1402, opened 2024-04-02
Add sorting to users table

syng828:
Added user sorting for issue #1389 
Sorts based on column name with buttons for each column. There are 3 states 'desc', 'asc', and 'unused', and it defaults to sorting by joinDate in descending order.

- Altered /routes/User to grab the queries and sort by the query
- Altered /APIFunctions/User to include sort and order parameters in request
- Altered /Overview to contain states for column, order, and states.

### DIFF
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -187,8 +187,8 @@ router.post('/users', async function(req, res) {
     desc: -1,
     asc: 1,
     default: -1
-  }
-  const sortOrder = orderToInteger[req.query.order] || orderToInteger.default
+  };
+  const sortOrder = orderToInteger[req.query.order] || orderToInteger.default;
 
   // make sure that the page we want to see is 0 by default
   // and avoid negative page numbers
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -181,13 +181,16 @@ router.post('/users', async function(req, res) {
     };
   }
 
+  const sortColumn = req.query.sort || 'joinDate';
+  const sortOrder = req.query.order === 'desc' ? -1 : (req.query.order === 'asc' ? 1 : -1);
+
   // make sure that the page we want to see is 0 by default
   // and avoid negative page numbers
   let skip = Math.max(Number(req.body.page) || 0, 0);
   skip *= ROWS_PER_PAGE;
   const total = await User.count(maybeOr);
   User.find(maybeOr, { password: 0, }, { skip, limit: ROWS_PER_PAGE, })
-    .sort({ joinDate: -1 })
+    .sort({ [sortColumn] : sortOrder })
     .then(items => {
       res.status(OK).send({ items, total, rowsPerPage: ROWS_PER_PAGE, });
     })
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -182,7 +182,13 @@ router.post('/users', async function(req, res) {
   }
 
   const sortColumn = req.query.sort || 'joinDate';
-  const sortOrder = req.query.order === 'desc' ? -1 : (req.query.order === 'asc' ? 1 : -1);
+
+  const orderToInteger = {
+    desc: -1,
+    asc: 1,
+    default: -1
+  }
+  const sortOrder = orderToInteger[req.query.order] || orderToInteger.default
 
   // make sure that the page we want to see is 0 by default
   // and avoid negative page numbers
--- a/src/APIFunctions/User.js
+++ b/src/APIFunctions/User.js
@@ -11,11 +11,27 @@ let GENERAL_API_URL = process.env.REACT_APP_GENERAL_API_URL
  * @returns {UserApiResponse} Containing any error information or the array of
  * users.
  */
-export async function getAllUsers(token, query = null, page = null, sortColumn = null, sortOrder = null) {
+export async function getAllUsers({
+  token,
+  query = null,
+  page = null,
+  sortColumn = null,
+  sortOrder = null,
+}) {
+  const url = new URL(GENERAL_API_URL + '/User/users')
+
+  if (sortColumn) {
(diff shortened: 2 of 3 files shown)


api/main_endpoints/routes/User.js
```diff
@@ -181,13 +181,16 @@ router.post('/users', async function(req, res) {
     };
   }
 
+  const sortColumn = req.query.sort || 'joinDate';
+  const sortOrder = req.query.order === 'desc' ? -1 : (req.query.order === 'asc' ? 1 : -1);
```
evanugarte: lets add a simple dictionary like
```js
const orderToInteger = {
  desc: -1
  asc: 1,
  default
}
const sortOrder = orderToInteger[req.query.order] || orderToInteger.default
```

src/APIFunctions/User.js
```diff
@@ -11,11 +11,11 @@ let GENERAL_API_URL = process.env.REACT_APP_GENERAL_API_URL
  * @returns {UserApiResponse} Containing any error information or the array of
  * users.
  */
-export async function getAllUsers(token, query = null, page = null) {
+export async function getAllUsers(token, query = null, page = null, sortColumn = null, sortOrder = null) {
```
evanugarte: lets turn this function into accepting an object, that way the order of the params dont matter
```suggestion
export async function getAllUsers({
    token,
    query = null,
    page = null,
    sortColumn = null,
    sortOrder = null,
}) {
```

src/APIFunctions/User.js
```diff
@@ -11,11 +11,11 @@ let GENERAL_API_URL = process.env.REACT_APP_GENERAL_API_URL
...
   let status = new UserApiResponse();
   await axios
     // get all user!
-    .post(GENERAL_API_URL + '/User/users', {
+    .post(GENERAL_API_URL + '/User/users' + `?sort=${sortColumn}&order=${sortOrder}`, {
```
evanugarte: we can use the `URL` class, its build into js like

```js
const url = new URL(GENERAL_API_URL + '/User/users')

if (sortColumn) {
  url.searchParams.set('sort', sortColumn);
}

if (sortOrder) {
  url.searchParams.set('order', sortOrder);
}

await axios
    .post(url.href, {
// ...
```

api/main_endpoints/routes/User.js
syng828: resolved with [changed to dict]

src/APIFunctions/User.js
syng828: resolved with @syng828
[changed getAllUsers function into accepting object + use url class to…](https://github.com/SCE-Development/Clark/pull/1402/commits/2761592fda73f52b539f4f3796b2df6acac05fe6)

src/APIFunctions/User.js
syng828: resolved with @syng828
[changed getAllUsers function into accepting object + use url class to…](https://github.com/SCE-Development/Clark/pull/1402/commits/2761592fda73f52b539f4f3796b2df6acac05fe6)

evanugarte (commented):
what if our sorting logic in the frontend was only handled by two hooks, "currentSortColumn" and "currentSortOrder"

- we modify the array that stores each column to have a new field, columnName
- when we map this array, we call handleSortUsers, passing in the columnName field to the function
- in handleSortUsers, we check the currentSortColumn value with the column name. if they match, then we either change the sort order from asc -> desc or desc -> none
- if they don't match, we set currentSortColumn to the column name and currentSortOrder to asc

src/Pages/Overview/Overview.js
```diff
@@ -219,17 +244,17 @@ export default function Overview(props) {
...
-                  { title: 'Verified', className: 'text-base text-white/70 text-center hidden sm:table-cell' },
-                  { title: 'Membership', className: 'text-base text-white/70 hidden text-center sm:table-cell' },
-                  { title: 'Delete', className: 'text-base text-white/70 text-center' },
-                ].map(({ title, className }) => (
+                  { title: 'Name/Email', className: 'text-base text-white/70', columnName: 'email'},
+                  { title: 'Printing', className: 'text-base text-white/70 hidden text-center md:table-cell', columnName: 'pagesPrinted'},
+                  { title: 'Verified', className: 'text-base text-white/70 text-center hidden sm:table-cell', columnName: 'emailVerified'},
+                  { title: 'Membership', className: 'text-base text-white/70 hidden text-center sm:table-cell', columnName: 'accessLevel'},
+                  { title: 'Delete', className: 'text-base text-white/70 text-center', columnName: 'delete'},
+                ].map(({ title, className, columnName }) => (
```
evanugarte: for columns we dont want to sort by, skip adding the `columnName` field. we can just default to null here
```suggestion
                ].map(({ title, className, columnName = null }) => (
```

src/Pages/Overview/Overview.js
```diff
@@ -85,6 +95,21 @@ export default function Overview(props) {
     );
   }, [page, rowsPerPage, users, total]);
 
+  function handleSortUsers(columnName) {
```
evanugarte, line 98: we need to add
```js
if (columnName === null) {
  return;
}
```

after the other coment about default null fields

src/Pages/Overview/Overview.js
syng828: resolved w/recent commit

src/Pages/Overview/Overview.js
syng828: resolved w/recent commit

src/Pages/Overview/Overview.js
```diff
@@ -85,6 +95,24 @@ export default function Overview(props) {
...
+    if (columnName == null) {
```
evanugarte: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness
```suggestion
    if (columnName === null) {
```

evanugarte (approved):
looks good

weslayer (approved):
nice

PR #1405, opened 2024-04-06
added url sorting

### LINKED ISSUE #1404: Url Sorting
Now that sorting parameters are added to Cleezy, implement sorting by columns in Clark.

syng828:
Resolves #1404

### DIFF
--- a/api/peripheral_api/routes/Cleezy.js
+++ b/api/peripheral_api/routes/Cleezy.js
@@ -27,7 +27,7 @@ router.get('/list', async (req, res) => {
     });
   }
   const token = req.query.token;
-  const { page = 0, search } = req.query;
+  const { page = 0, search, sortColumn = "created_at", sortOrder = "DESC"} = req.query;
   if (!token) {
     return res.sendStatus(FORBIDDEN);
   } else if (!await verifyToken(req.query.token)) {
@@ -38,6 +38,8 @@ router.get('/list', async (req, res) => {
       params: {
         page,
         ...(search !== undefined && { search }),
+        sort_by: sortColumn,
+        order: sortOrder
       },
     });
     const { data = [], total, rows_per_page: rowsPerPage } = response.data;
--- a/api/peripheral_api/routes/Cleezy.js
+++ b/api/peripheral_api/routes/Cleezy.js
@@ -27,7 +27,7 @@ router.get('/list', async (req, res) => {
     });
   }
   const token = req.query.token;
-  const { page = 0, search, sortColumn = "created_at", sortOrder = "DESC"} = req.query;
+  const { page = 0, search, sortColumn = 'created_at', sortOrder = 'DESC'} = req.query;
   if (!token) {
     return res.sendStatus(FORBIDDEN);
   } else if (!await verifyToken(req.query.token)) {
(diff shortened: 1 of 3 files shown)


api/peripheral_api/routes/Cleezy.js
```diff
@@ -38,6 +38,9 @@ router.get('/list', async (req, res) => {
       params: {
         page,
         ...(search !== undefined && { search }),
+        // eslint-disable-next-line camelcase
+        sort_by: sortColumn,
```
evanugarte, line 42: this is the correct way to get around this in my opinion

src/Pages/URLShortener/URLShortener.js
```diff
@@ -107,6 +117,24 @@ export default function URLShortenerPage(props) {
...
 
+  function handleSortUrls(columnName) {
+    if (columnName === null) {
+      return;
+    }
+    if(currentSortColumn === columnName) {
+      if (currentSortOrder === 'ASC') {
+        setCurrentSortOrder('DESC');
+      } else if (currentSortOrder === 'DESC') {
+        setCurrentSortOrder('none');
```
evanugarte: how about instead of `'none'` we just set to `null`?

src/Pages/URLShortener/URLShortener.js
```diff
@@ -33,9 +35,17 @@ export default function URLShortenerPage(props) {
    * Cleezy page is disabled by default since you have to run the Cleezy server
    * separately. To enable, go to config.json and set ENABLED under Cleezy to true
    */
-  async function getCleezyUrls(page, searchQuery) {
+  async function getCleezyUrls(page, searchQuery, currentSortColumn, currentSortOrder) {
     setLoading(true);
-    const urlsFromDb = await getAllUrls(props.user.token, page, searchQuery);
+    const sortColumn = currentSortOrder === 'none' ? 'created_at' : currentSortColumn;
+    const sortOrder = currentSortOrder === 'none' ? 'DESC' : currentSortOrder;
```
evanugarte: see the null comment then we can do

```suggestion
    const sortColumn = currentSortColumn ?? 'created_at';
    const sortOrder = currentSortOrder ?? 'DESC';
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/Pages/URLShortener/URLShortener.js
syng828: resolved, changed logic to set column to null as well

src/Pages/URLShortener/URLShortener.js
```diff
@@ -25,6 +25,8 @@ export default function URLShortenerPage(props) {
   const [errorAlertMessage, setErrorAlertMessage] = useState('');
   const [urlToDelete, setUrlToDelete] = useState({});
   const [toggleDelete, setToggleDelete] = useState(false);
+  const [currentSortColumn, setCurrentSortColumn] = useState('created_at');
+  const [currentSortOrder, setCurrentSortOrder] = useState('DESC');
```
evanugarte: initialize these as null, `getCleezyUrls` has logic to use default values

evanugarte (approved):

weslayer (approved):
lgtm

PR #1407, opened 2024-04-16
Fix the majors available on the registration page and on edit users.

weslayer:
Fix evans silly mistake heehehaha


evanugarte (approved):
how about for the edit user page?

syng828 (approved):
looks good

PR #1442, opened 2024-06-26
added qr to nginx

syng828:
Added qr to nginx for qr code generation from cleezy


evanugarte (approved):
total syng828 victory

weslayer (approved):
swag alert is going off

PR #1443, opened 2024-06-27
fixed membership form issue

lac-phong:
fixed membership form so that user is only told the username is taken if the response is a 409 error and also deleted unnecessary function sendVerificationEmail


evanugarte (approved):

flotoria (approved):
lgtm

PR #1454, opened 2024-07-08
Update user's print count in backend

joshuakojko:
Subtracts user's print count in /sendPrintRequest 
issue #1004

### DIFF
--- a/api/peripheral_api/routes/Printer.js
+++ b/api/peripheral_api/routes/Printer.js
@@ -4,32 +4,26 @@ const logger = require('../../util/logger');
 const {
   verifyToken,
   checkIfTokenSent,
-} = require('../../util/token-verification');
-const {
-  OK,
-  UNAUTHORIZED,
-  NOT_FOUND,
-  SERVER_ERROR,
-  BAD_REQUEST,
-} = require('../../util/constants').STATUS_CODES;
-const {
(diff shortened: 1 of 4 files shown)


api/peripheral_api/routes/Printer.js
```diff
@@ -10,11 +10,12 @@ const {
   UNAUTHORIZED,
   NOT_FOUND,
   SERVER_ERROR,
+  BAD_REQUEST,
 } = require('../../util/constants').STATUS_CODES;
 const {
   PRINTING = {}
 } = require('../../config/config.json');
-
+const User = require('../../main_endpoints/models/User')
```
evanugarte: the way the code is set up right now, we cant access the user model from the `peripheral_api` folder. this is my fault because of https://github.com/SCE-Development/Clark/pull/491, 

we need to call on the edit user api with axios to update pages printed, 

similar to below, let me know if this makes sense

https://github.com/SCE-Development/Clark/blob/053254e98c69909dd5e834cbb17fc4592a9c63fb/api/peripheral_api/routes/LedSign.js#L39

https://github.com/SCE-Development/Clark/blob/053254e98c69909dd5e834cbb17fc4592a9c63fb/api/util/token-verification.js#L15

PR #1455, opened 2024-07-08
chore: Add prom-client and express-prom-bundle dependencies

knhn1004:


api/util/SceHttpServer.js
```diff
@@ -7,6 +7,17 @@ mongoose.Promise = require('bluebird');
 
 const { PathParser } = require('./PathParser');
 const logger = require('./logger');
+const promBundle = require('express-prom-bundle');
```
evanugarte: can we use this library instead? https://github.com/siimon/prom-client

here is an example on how to use it https://github.com/evanugarte/prometheus-monitoring-tutorial/blob/6a2d98b52c43ea27e23c4e39ecd14a9f56c24d73/server.js#L54

PR #1456, opened 2024-07-08
Messaging API

jonathanguven:
Frontend for the Messaging API Platform

1. Form for sending messages
2. Live feed for default room


src/APIFunctions/Messaging.js
```diff
@@ -0,0 +1,36 @@
...
+    })
+    .catch(err => {
+      status.error = true;
+      status.responseData = err;
+    });
+  return status;
+}
+
+export async function connectToRoom(room, apiKey, onMessage, onError) {
+  const eventSource = new EventSource(`${GENERAL_API_URL}/messages/listen?id=${encodeURIComponent(room)}&apiKey=${encodeURIComponent(apiKey)}`);
```
evanugarte: can we do this, also on line 11 too
```suggestion
  const url = new URL(`${GENERAL_API_URL}/messages/listen`);
  url.searchParams.append('id', room);
  url.searchParams.append('apiKey', apiKey);
  const eventSource = new EventSource(url.href);
```

src/APIFunctions/Messaging.js
```diff
@@ -0,0 +1,36 @@
...
+
+  eventSource.onmessage = (event) => {
+    const data = JSON.parse(event.data);
```
evanugarte: add a try/catch in here? or will `eventSource.onerror` handle paring errors for us?

src/Pages/Messaging/Messaging.js
```diff
@@ -0,0 +1,169 @@
...
+function Feed(props) {
+  const { token } = props;
+  const [messages, setMessages] = useState([]);
+  const [apiKey, setApiKey] = useState('');
+  const [error, setError] = useState('');
+  const roomId = 'general';
+  let eventSource;
+
+  useEffect(() => {
+    const fetchApiKey = async () => {
```
evanugarte: how about for sending messages, we allow the user to pass in a token, that way we dont need to handle api key logic in here


api keys can be the second option that our messaging apis use, if a token wasnt sent, but an api key was, let me know if this makes sense

by allowing tokens, we dont need this function

src/APIFunctions/Messaging.js
flotoria: i think ```eventSource.onerror``` should handle the errors

src/Pages/Messaging/Messaging.js
flotoria: lemme handle this

flotoria: implemented, probably need to check whether the solution is clean or not

akshtshrma24 (approved):

akshtshrma24 (approved):

akshtshrma24 (approved):

flotoria (approved):

evanugarte (approved):

food0903 (approved):

PR #1458, opened 2024-07-08
20240707-prometheus-client-setup

knhn1004:


knhn1004:
it seems to be breaking some endpoints' creation, requires further testing!

knhn1004:
refactoring handler substitution method to be using proper middlewares

PR #1461, opened 2024-07-09
got JWT token to be passed through the authorization header

lac-phong:


src/APIFunctions/Auth.js
```diff
@@ -93,7 +92,11 @@ export async function checkIfUserIsSignedIn() {
   }
 
   await axios
-    .post(GENERAL_API_URL + '/Auth/verify', { token })
+    .post(GENERAL_API_URL + '/Auth/verify', {}, {
+      headers: {
+        Authorization: `${token}`
```
evanugarte: make the header look like `Bearer <token>`, this follows a convention
https://stackoverflow.com/a/33281233

PR #1469, opened 2024-07-12
added random text on top of homepage from api response

cuzethan:
@evanugarte followed the discord instructions


api/main_endpoints/routes/HomePageText.js
```diff
@@ -0,0 +1,10 @@
+const express = require('express');
+const router = express.Router();
+
+router.get('/getText', (req, res) => {
```
evanugarte: can we put this in the below file? we can remove the other endpoints too
https://github.com/SCE-Development/Clark/blob/06b0807916a0a2b62d13256607dcce61b3edd594/api/main_endpoints/routes/Advertisement.js#L5

cuzethan: @evanugarte resolved, but the error that appears seems to come from the endpoints that you told me to delete

evanugarte:
#1489, #1490 and #1494 resolved this

PR #1472, opened 2024-07-17
Initial implementation volume API

stanphong:
Add new volume function in speaker.js in api function folder and speaker.js in Speaker folder
Volume bar works in front-end and sends the value back to the API
API does not actually work yet.

image


stanphong:
Update front-end with forward, reverse control
image

src/Pages/Speaker/Speaker.js
```diff
@@ -34,24 +38,57 @@ function SpeakersPage(props) {
...
+    } else {
+      await modifySpeakerWrapper(resume);
+    }
+    setIsPlaying(!isPlaying);
+  };
+
+  const handleVolumeChange = async (volume) => {
+    setVolumeState(volume); // Update local volume state
+    const result = await setVolume(volume, props.user.token);
+  };
```
evanugarte: will this get called any time the slider changes value, or when the user is done dragging the slider?

akshtshrma24: When user is done dragging slider

akshtshrma24 (approved):

src/Pages/Speaker/Speaker.js
akshtshrma24: now it slike theat

akshtshrma24 (approved):

evanugarte (approved):

PR #1473, opened 2024-07-17
Move JWT token to HTTP Auth Header

lac-phong:


pnavab (approved):

evanugarte (approved):
tested and it seems like it works

PR #1475, opened 2024-07-19
Add back routing tests

armanbance:
#171 all over again


evanugarte (approved):
we are getting warnings about components changing state after they render, I will resolve that in a separate pr

weslayer (approved):
aprove

src/Routing.js
```diff
@@ -31,6 +31,7 @@ import EmailPreferencesPage from './Pages/EmailPreferences/EmailPreferences';
 import sendUnsubscribeEmail from './Pages/Profile/admin/SendUnsubscribeEmail';
 import Messaging from './Pages/Messaging/Messaging.js';
 
+
```
evanugarte:
```suggestion
```

PR #1508, opened 2024-08-13
Messaging API timestamps

flotoria:
- Backend wise, JSONify the messages so the messages broadcasted include timestamps (broadcasted in a JSON string) 
- Have the message and its corresponding timestamp displayed on the frontend

### DIFF
--- a/api/main_endpoints/routes/Messages.js
+++ b/api/main_endpoints/routes/Messages.js
@@ -34,7 +34,7 @@ const writeMessage = ((roomId, message) => {
           hour12: false
         }
       ),
-    message: message
+    message,
   };
 
   if (clients[roomId]) {
--- a/api/main_endpoints/routes/Messages.js
+++ b/api/main_endpoints/routes/Messages.js
@@ -20,10 +20,19 @@ const numberOfConnections = {};
 const lastMessageSent = {};
 
 const writeMessage = ((roomId, message) => {
+
+  const currentTimestamp = new Date();
+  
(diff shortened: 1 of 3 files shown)


food0903 (approved):

api/main_endpoints/routes/Messages.js
```diff
@@ -20,10 +20,28 @@ const numberOfConnections = {};
...
+      currentTimestamp.toLocaleDateString() + ' ' +
+      currentTimestamp.toLocaleTimeString(
+        [],
+        {
+          hour: '2-digit',
+          minute: '2-digit',
+          hour12: false
+        }
+      ),
+    message: message
```
evanugarte:
```suggestion
    message,
```

api/main_endpoints/routes/Messages.js
```diff
@@ -20,10 +20,28 @@ const numberOfConnections = {};
...
+    timestamp:
+      currentTimestamp.toLocaleDateString() + ' ' +
+      currentTimestamp.toLocaleTimeString(
+        [],
+        {
+          hour: '2-digit',
+          minute: '2-digit',
+          hour12: false
+        }
+      ),
```
evanugarte: how about we return the epoch timetamp in ms, and the frontend does the parsing of it

```suggestion
    timestamp: Date.now(),
```

evanugarte (commented):
how about

akshtshrma24 (approved):

PR #1509, opened 2024-08-14
Add Prometheus metrics to Messaging API

flotoria:
- Initialized Prometheus metrics for Messaging API 
- Added the following metrics that Prometheus will monitor: current_rooms_open, total_messages_sent, current_connections_open, total_chat_messages_per_chatroom, and total_rooms_opened

### DIFF
--- a/api/main_endpoints/routes/Messages.js
+++ b/api/main_endpoints/routes/Messages.js
@@ -10,18 +10,61 @@ const router = express.Router();
 const bodyParser = require('body-parser');
 const User = require('../models/User.js');
 const logger = require('../../util/logger');
+const client = require('prom-client');
 const { decodeToken, decodeTokenFromBodyOrQuery } = require('../util/token-functions.js');
 
 
 router.use(bodyParser.json());
+
+// all data collected about server activity
+let register = new client.Registry();
+
+// all the metrics for the messages api
+const currentRoomsOpen = new client.Gauge({
+  name: 'current_rooms_open',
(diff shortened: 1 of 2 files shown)


evanugarte (commented):
can we move the metrics into the below file
https://github.com/SCE-Development/Clark/blob/6b56da1b1a3f896b2a15137a57e50f21e23773cf/api/util/metrics.js#L22

an example of importing and using the metrics is here
https://github.com/SCE-Development/Clark/blob/6b56da1b1a3f896b2a15137a57e50f21e23773cf/api/main_endpoints/util/captcha.js#L22

akshtshrma24 (commented):
add a last message sent

api/main_endpoints/routes/Messages.js
```diff
@@ -180,17 +188,30 @@ router.get('/listen', async (req, res) => {
...
+
+        // increase total amount of rooms opened
+        MetricsHandler.totalRoomsOpened.inc();
+
+        // increase the current rooms open gauge
+        MetricsHandler.currentRoomsOpen.inc();
       }
 
+      // add connection to the connections open gauge
+      MetricsHandler.currentConnectionsOpen.inc();
```
evanugarte: how about instead of current and total rooms open, we just use this metric

`currentConnectionsOpen` can have a label, `roomId` so its easy to know how many connections are open and for what room

food0903 (approved):
lgtm!

trungtran1234 (approved):
good

Issue #1555, opened 2025-01-12
Remove unconditional query to mongodb on each chat message send


evanugarte:
right now, every time we send a chat message, it results in a query to read from mongodb to retrieve the users first name, see
https://github.com/SCE-Development/Clark/blob/5f89a180a05096be25033d028a5b1d6d45a1cddf/api/main_endpoints/routes/Messages.js#L79

We can skip querying the database everytime. instead, lets change the code to do this

- [ ] change the frontend to pass in the first name of the user as part of the request when sending a message (see `props.user`)
- [ ] if the user specified an api key, do the mongodb query to find the user's first name based on their api key, like we do now
- [ ] else, if the user didnt send an api key, use the first name sent in as a parameter of the request

PR #1564, opened 2025-01-15
Remove q

PatrickTle924:
Followed the instructions in setting up the Messaging Component and the Form to send the username in the request. The request is then passed through and used in the endpoint to avoid querying the DB if user is not using an APIKey. Apologies if my Prettier messed up the formatting.

concering issue #1555

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -69,10 +69,7 @@ http {
         #Load balancer
         location /api {
             proxy_pass http://mainendpoints;
-        }
-        location /spark {
-            proxy_pass https://spark-scesjsu.webflow.io/;
-        }
+        
         location / {
             proxy_pass http://webserver;
         }


api/main_endpoints/routes/Messages.js
```diff
@@ -1,38 +1,36 @@
...
-    username
+    username,
   };
 
   if (clients[roomId]) {
-    clients[roomId].forEach(res => res.write(`data: ${JSON.stringify(messageObj)}\n\n`));
+    clients[roomId].forEach((res) =>
+      res.write(`data: ${JSON.stringify(messageObj)}\n\n`)
+    );
   }
```
evanugarte: there are a bunch of random changes in this file. lets disable prettier on vscode, and undo these changes

the only diff in this file should be the logic to remove the mongodb query

PatrickTle924:
Sorry about the Prettier formatting issues. Looking over the change log I can see it caused a ton of unnecessary changes that probably makes reviewing the PR pretty tough. I rolled back the commits, disabled Prettier, and pushed a new commit that should have the changes cleaner, and without the random formatting changes.

api/main_endpoints/routes/Messages.js
```diff
@@ -63,34 +63,36 @@ router.post('/send', async (req, res) => {
...
+        }
+        if (result) {
+          writeMessage(id, `${message}`, `${result.firstName}:`);
+          return res.json({ status: 'Message sent' });
+        }
+        return res.sendStatus(UNAUTHORIZED);
+      });
+    } catch (error) {
+      logger.error('Error in /send: ', error);
+      res.sendStatus(SERVER_ERROR);
```
evanugarte:
```suggestion
      return res.sendStatus(SERVER_ERROR);
```

api/main_endpoints/routes/Messages.js
```diff
@@ -63,34 +63,36 @@ router.post('/send', async (req, res) => {
...
     }
-    filterQuery._id = userObj._id;
   } else {
```
evanugarte: we can remove the entire `else` indentation, if we use `return` statements inside `if (apiKey) {`

let me know if this makes sense. it makes the code easier to follow

```js
if (thing) {
  // stuff
  return idk;
}
// no more else block
// just the original code that wouldve been in the else ere
```

PatrickTle924: Understood, removed the else block. I had to change the code to make the User.findOne() function call async since the code ended up sending multiple responses (I think it was running the code after the if (apiKey) block as it waited for the User.findOne() function?). Please let me know if that's okay or there was an alternative I should have done. Thanks!

NicholasLe04 (approved):
lgtm

api/main_endpoints/routes/Messages.js
```diff
@@ -64,34 +63,30 @@ router.post('/send', async (req, res) => {
     return;
   }
 
-  let filterQuery = {}; // filter to find user in the database
+  if (apiKey) {
+    let filterQuery = {};
+    filterQuery.apiKey = apiKey;
+    try {
+      const result = await User.findOne(filterQuery).exec();
```
evanugarte:
```suggestion
    try {
      const result = await User.findOne({ apiKey }).exec();
```

evanugarte (approved):
was this tested

api/main_endpoints/routes/Messages.js
PatrickTle924: Yessir, I tested it using Postman and creating a request with the x-api-key in the header. I will make the changes suggested. Thank you!

PR #1568, opened 2025-01-20
1553 persist chat

ngtri1809:


api/main_endpoints/models/ChatMessage.js
```diff
@@ -0,0 +1,32 @@
...
+      type: String,
+      required: true,
+    },
+    createdAt: {
+      type: Date,
+      default: Date.now(),
+    },
+    expireAt: {
+      type: Date,
+      default: Date.now() + 10 * 60 * 1000 // expires in 10 minutes
```
evanugarte: should the expiration be in a day? let's instead use the `createdAt` field to calculate when it expires

api/main_endpoints/models/ChatMessage.js
```diff
@@ -0,0 +1,32 @@
...
+const ChatMessageSchema = new Schema(
+  {
+    userId: {
+      type: String,
+      required: true,
+    },
+    text: {
+      type: String,
+      required: true,
+      maxLength: [255, 'Messages are at most 255 characters long']
```
evanugarte: how about 2000 characters instead

api/main_endpoints/models/ChatMessage.js
ngtri1809: Roger!

api/main_endpoints/models/ChatMessage.js
ngtri1809: Oh yeah that sounds great xD.

PR #1578, opened 2025-02-03
Fix: Matched Textbox and Text Color on URL Shortener Page

alanvu2440:
- Fixed the color of the textbox and text on the URL Shortener page when you click on "Create a new link", so it is now visible in light mode, and matches the bottom textbox and text color. 
Screenshot 2025-02-03 at 10 16 13 PM


weslayer (approved):
swag

evanugarte (approved):

PR #1586, opened 2025-02-09
Un-nest user registration logic

evanugarte:
makes the logic easier to follow and resolves using both async and `.then` syntax in the same place


weslayer (approved):
so swag evan ugarte

weslayer (approved):
aprove

NicholasLe04 (approved):
fire

PR #1602, opened 2025-03-04
SCEats Admin Dashboard

brendan-ly:
Implemented the admin frontend dashboard for SCEats to view, manage, and delete snacks for SCE's snack inventory. 
Screenshot 2025-03-03 at 5 24 47 PM


api/food_api/routes/Food.js
```diff
@@ -0,0 +1,90 @@
...
+  OK,
+  BAD_REQUEST,
+  NOT_FOUND
+} = require('../../util/constants').STATUS_CODES;
+
+router.post('/createFood', (req, res) => {
+  const { price, quantity } = req.body;
+  const isValidNumber = (value) => {
+    return !Number.isNaN(Number(value));
+  }
```
al3xbro: this seems like a great utility function! why don't we move it to a separate utils folder so other files can use it, and import it here?

api/food_api/routes/Food.js
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  Food.create(newFood)
+    .then((post) => {
+      return res.json(post);
+    })
+    .catch(
+      (error) => res.sendStatus(BAD_REQUEST)
+    );
+});
```
al3xbro: `.then()` syntax is hard to read when lengthy, and even tougher to use in complex code. the preferred way of handling promises is almost always async await: https://pusher.com/blog/promises-async-await/#conditionals

api/food_api/routes/Food.js
```diff
@@ -0,0 +1,90 @@
...
+        .then(() => {
+          res.sendStatus(OK);
+        })
+        .catch(() => {
+          res.sendStatus(BAD_REQUEST);
+        });
+    })
+    .catch(() => {
+      res.sendStatus(NOT_FOUND);
+    });
```
al3xbro: an example of how `.then()` can get messy, and this is only 2 `.then()`s

src/Pages/SCEatsAdmin/FoodForm.js
```diff
@@ -0,0 +1,77 @@
...
+            </div>
+            <input type="text" maxLength={50} placeholder="For example, Cheetos" className="input input-bordered w-full max-w-md" onChange={e => setName(e.target.value)} />
+          </label>
+        </div>
+
+        <div className="flex-1 min-w-[200px]">
+          <div className="label">
+            <span className="label-text">Type</span>
+          </div>
+          <select onChange= {e => setType(e.target.value)} className="select select-bordered w-full max-w-md">
```
al3xbro, line 22: no `value={}`?

src/Pages/SCEatsAdmin/FoodGrid.js
```diff
@@ -0,0 +1,149 @@
...
+                  className="rounded-md bg-indigo-600 mt-1 px-4 py-2 text-sm font-semibold text-white shadow-sm hover:bg-indigo-500 focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-indigo-600"
+                  onClick={() => handleSave()}>Save</button>
+              ) : (
+                <>
+                  <button onClick={() => handleEdit(food)} className="bg-blue-500 text-white px-2 py-2 rounded-md">
+                    <svg xmlns="http://www.w3.org/2000/svg" width="1.2em" height="1.2em" viewBox="0 0 24 24"><g fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"><path stroke-dasharray="20" stroke-dashoffset="20" d="M3 21h18"><animate fill="freeze" attributeName="stroke-dashoffset" dur="0.2s" values="20;0" /></path><path stroke-dasharray="48" stroke-dashoffset="48" d="M7 17v-4l10 -10l4 4l-10 10h-4"><animate fill="freeze" attributeName="stroke-dashoffset" begin="0.2s" dur="0.6s" values="48;0" /></path><path stroke-dasharray="8" stroke-dashoffset="8" d="M14 6l4 4"><animate fill="freeze" attributeName="stroke-dashoffset" begin="0.8s" dur="0.2s" values="8;0" /></path></g></svg>
```
al3xbro: this should be in a separate folder in a separate file. it is extremely messy here

src/Pages/SCEatsAdmin/SCEatsAdmin.js
```diff
@@ -0,0 +1,242 @@
...
+
+  const [editingId, setEditingId] = useState(null);
+  const [editedFood, setEditedFood] = useState({
+    name: '',
+    type: '',
+    photo: '',
+    price: '',
+    quantity: '',
+    expiration: ''
+  });
```
al3xbro: this is a crazy amount of states. i don't think most of these need to be in the same component. i haven't read all the code, but if it needs to be, then the component was designed poorly. consider compartmentalizing them

src/Pages/SCEatsAdmin/SCEatsAdmin.js
al3xbro, line 1: this file is waaayyyy too big. consider placing small items (like buttons and their functions) into separate files, and reusing them instead of copy pasting html. most of the time, copy pasting code means you're violating [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). 

react calls for [atomic design pattern](http://medium.com/@janelle.wg/atomic-design-pattern-how-to-structure-your-react-application-2bb4d9ca5f97) (that's why the logo is an atom). since that would require way too many changes, let's ignore that, but keep the spirit of atomic design alive in our code. just try splitting up these files. most react components should not be more than 100 lines

al3xbro (changes requested):
great job on the code guys. make sure you address each comment and apply it to other parts of the code as well. i mentioned each issue at least once, but didn't highlight every single occurance

PR #1611, opened 2025-03-09
use originalUrl field in prometheus route label for api requests

evanugarte:
now we get the full path of the request, like
```
sce-main-endpoints      | /api/Auth/login
sce-main-endpoints      | /api/User/edit
sce-main-endpoints      | /api/Advertisement/
sce-main-endpoints      | /api/Auth/verify
sce-main-endpoints      | /api/Advertisement/
```

before it was below, which doesnt give much context as we dont know what the full path of the request was

```
sce-main-endpoints      | login
sce-main-endpoints      | edit
sce-main-endpoints      | /
sce-main-endpoints      | verify
sce-main-endpoints      | /
```

PR #1620, opened 2025-03-15
use req._parsedUrl.pathname in prometheus route label for api requests

evanugarte:
#1611 worked, but it was including query parameters in the route labels like `/api/something?key=value`

this removes the query parameters from the label so we get consistent labels across api requests, i.e. `/api/something`


gursheyss (approved):
epic

n8thantran (approved):
lgtm

PR #1639, opened 2025-03-26
Verification count and last verified date

mceballos123:
Created a feature that increments by one whenever a card is verified and checks the last time the card was verified. This allows SCE to check which office access cards are active and which are inactive


akshtshrma24 (approved):

M1Z8N (approved):
lgtm

PR #1648, opened 2025-04-10
Add hyperlinks to club announcement messages (#1569)

### LINKED ISSUE #1569: Add hyperlinks into the club announcement messages
on the homepage, messages are always displayed as plaintext, even urls.

Image

lets add some logic in the frontend to dynamically add a blue color and underlining to parts of the announcement string that are valid urls. clicking on a url should open the link in a new tab.

M1Z8N:
# Changes
Formats URLs in club announcements as clickable links
Links are underlined, blue, and open in a new tab

# Screenshots

**Before:**
before

**After:**
after
Closes #1569

### DIFF
--- a/src/Pages/Home/Home.js
+++ b/src/Pages/Home/Home.js
@@ -24,9 +24,21 @@ const Home = () => {
     getMessage();
   }, []);
 
+  function isValidUrl(str) {
+  try {
+    new URL(str);
+    return true;
+  } catch (_) {
+    return false;
+  }
+}
+
   function renderMessageWithLinks(message) {
+    if (!message) {
+      return null;
+    }
     return message.split(/(https?:\/\/[^\s]+)/g).map((part, index) => {
-      if (/https?:\/\/[^\s]+/.test(part)) {
+      if (isValidUrl(part)) {
         return (
           <a
             key={index}
--- a/src/Pages/Home/Home.js
+++ b/src/Pages/Home/Home.js
@@ -9,7 +9,6 @@ const Home = () => {
 
   const [message, setMessage] = useState('');
   const [showMessage, setShowMessage] = useState(false);
-
   async function getMessage() {
     try {
       const messageData = await getAd();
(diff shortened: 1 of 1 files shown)


hunter-nguyen (approved):

src/Pages/Home/Home.js
```diff
@@ -35,7 +35,21 @@ const Home = () => {
           className="text-white"
           style={{ display: showMessage ? 'block' : 'none' }} // Conditional display based on state
         >
-          {message}
+          {message.split(/(https?:\/\/[^\s]+)/g).map((part, index) =>
```
evanugarte: can we make this a function instead, the function can return some html

like
```js
function maybeInsertHyperlinkInAnnouncement(announcementString) {
  // split up the string, and then return a <p> tag,
  // with optional <a> tag inside
}
```

weslayer (approved):
regex moment

evanugarte (changes requested):
requesting changes to block merge

src/Pages/Home/Home.js
M1Z8N: What would be the best location for this function, keep it in Home.js?

evanugarte: that sounds good, underneath the `async function getMessage() {` function

M1Z8N:
Awesome Sauce

src/Pages/Home/Home.js
```diff
@@ -25,6 +25,24 @@ const Home = () => {
...
+          target="_blank"
+          rel="noopener noreferrer"
+          className="text-blue-400 underline"
+        >
+          {part}
+        </a>
+      ) : (
+        <span key={index}>{part}</span>
+      )
+    );
```
evanugarte: inside the `.map` function can we do something like
```js
if (https?:\/\/[^\s]+/.test(part)) {
  return // insert the <a ...
}
return <span key={index}>{part}</span>
```

evanugarte (approved):

PR #1673, opened 2025-06-07
remove speaker page

### LINKED ISSUE #1662: Remove lodash
lodash takes up 7.4% of bundle size but is only used in one file[`Speaker.js`] (https://github.com/SCE-Development/Clark/blob/dev/src/Pages/Speaker/Speaker.js), can be optimized to just use react hooks and remove lodash

Image

altran03:
Closes #1662

lodash bundle size actually can't be resolved because it is in other packages. This pr comments it out of navbar and deletes lodash in Speaker.js.


evanugarte (approved):

weslayer (approved):
goodbye speaker 😢

Issue #1676, opened 2025-06-08
Remove axios, use only fetch and URL class


DanielTQuach:
fix misuse of chained promises

PR #1684, opened 2025-06-10
2DPrinting remove axios

DanielTQuach:
regarding #1676
removing axiosfor only 2DPrinting.js

### DIFF
--- a/src/APIFunctions/2DPrinting.js
+++ b/src/APIFunctions/2DPrinting.js
@@ -1,4 +1,3 @@
-import axios from 'axios';
 import {
   PrintApiResponse,
   ApiResponse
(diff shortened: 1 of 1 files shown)


src/APIFunctions/2DPrinting.js
```diff
@@ -78,18 +80,22 @@ export function parseRange(pages, maxPages) {
...
-    .then(response => {
-      status.responseData = response.data.message;
-    })
-    .catch(() => {
-      status.error = true;
+  try {
+    const res = await fetch(url.href, {
+      method: 'POST',
+      headers: {
+        'Authorization': `Bearer ${token}`
```
evanugarte, line 87: do we need 'Content-Type': 'multipart/form-data',?

evanugarte (approved):
try running quasar in parallel with clark, and see if the file makes it to quasar?

https://github.com/SCE-Development/Quasar

be sure to enable printing below

https://github.com/SCE-Development/Clark/blob/46801ff13fd8842b17e29fdd99094097800292da/api/config/config.example.json#L21

n8thantran (approved):
lgtm

Issue #1682, opened 2025-06-10
create test to verify that login succeeds with correct credentials


pxtrickhoxng:
The POST /login endpoint currently has unit tests that cover failed scenarios (returning 400 for missing fields and 401 for incorrect credentials), but doesn’t have a positive test case.

There should be a new test that verifies that the login works when provided with the correct credentials.

Because login requires an existing user, the test should **first** register a user, then attempt to log in with those same credentials. The response should return 200.

PR #1686, opened 2025-06-11
Add test verifying login succeeds with correct credentials

pxtrickhoxng:
This PR adds a test that first registers a user, sets their email as verified, and then attempts to log in using the same credentials. It verifies that the login succeeds and that the returned token contains the correct structure and user information.

Addresses issue #1682.

### DIFF
--- a/test/api/Auth.js
+++ b/test/api/Auth.js
@@ -210,14 +210,14 @@ describe('Auth', () => {
 
       const decodedPayload = decodeToken(mockRequest);
       const expectedPayload = {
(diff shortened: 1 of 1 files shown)


weslayer:
Make it so that the user is deleted from the DB after the test runs.

evanugarte (approved):

flotoria:
looks like the lint check failed, run `npx eslint . --fix` and commit again

weslayer (approved):

PR #1688, opened 2025-06-12
Club Semester Revenue API

Funsize1234:
Gets number of signups for the semester who've paid semester plan and number of those who've paid annual plan, number of current active members and number of new members this year. Right now doesn't have a way to track recurring members.


evanugarte (approved):

weslayer (approved):
nice

evanugarte:
see #1720

Issue #1695, opened 2025-06-12
Cleezy API functions remove axios


DanielTQuach:
Currently using chained promises along with axios. We can replace this with just fetch and await

PR #1696, opened 2025-06-12
Cleezy API functions replace axios

DanielTQuach:
#1695 
replacing axios + chained promises with just fetch and await


evanugarte (approved):

adarshm11 (approved):
nice work

Issue #1697, opened 2025-06-13
Remove axios for Mailer API Functions


DanielTQuach:
Currently using chained promises along with axios. We can replace this with just fetch and await

PR #1698, opened 2025-06-13
Mailer API functions replace axios

DanielTQuach:
#1697 
 replacing axios + chained promises with just fetch and await


evanugarte (approved):

thebeninator (approved):
lgtm

PR #1718, opened 2025-06-19
Reordered inputs of Cleezy URL Shortner

KingSand08:
- Reordered inputs (alias, toggle for alias, then original url)
- toggle default is toggled off (false)
- changed the url css to be light-mode compatible (originally only dark-mode)
image
Screenshot 2025-06-23 at 10 51 58 PM


evanugarte (approved):

n8thantran (approved):
this is awesome.

PR #1723, opened 2025-06-24
add userContext

altran03:
Start implementing useContext from React instead of props


evanugarte (approved):

weslayer (approved):
goodbye props, no one ever liked you

PR #1730, opened 2025-06-28
Implement audit log on user edit + tests

pxtrickhoxng:
Add an audit log for when a user is updated.

### DIFF
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -32,6 +32,9 @@ const crypto = require('crypto');
 
 const ROWS_PER_PAGE = 20;
 
+const AuditLogActions = require('../util/auditLogActions.js');
+const AuditLog = require('../models/AuditLog.js');
+
 // Delete a member
 router.post('/delete', async (req, res) => {
   if (!checkIfTokenSent(req)) {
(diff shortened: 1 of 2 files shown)


api/main_endpoints/routes/User.js
```diff
@@ -235,6 +238,14 @@ router.post('/edit', async (req, res) => {
         .status(NOT_FOUND)
         .send({ message: `${query.email} not found.` });
     }
+
+    AuditLog.create({
+      userId: decoded._id, // the user making the update
+      action: AuditLogActions.UPDATE_USER,
+      documentId: user._id, // the user affected by the update
+      details: {updatedInfo: user}
```
evanugarte: - edit a user, change their password
- does the password field appear in the AuditLog collection?
- if so, lets add logic to replace password to `true`, if it is set, so we avoid storing it

evanugarte: add a unit test to verify this as well

test/api/User.js
```diff
@@ -224,6 +227,63 @@ describe('User', () => {
...
+      // ensure Audit log DB starts fresh before this test
+      beforeEach(async () => {
+        await AuditLog.deleteMany({});
+      });
+
+      afterEach(async () => {
+        await AuditLog.deleteMany({});
+      });
+
+      it('Should create an audit log when a user is updated', async () => {
```
evanugarte: this could probably be one test? since they both test the same thing except for one extra assertion

evanugarte (approved):

evanugarte (approved):

weslayer (approved):

PR #1735, opened 2025-06-29
changed input box

leishaaf:
made changes to Advertisement.js (removed slashes) and changed the input box from 3 sep fields to one date


evanugarte:
we used this code in https://github.com/SCE-Development/Clark/pull/1760

PR #1742, opened 2025-07-01
PR3: Add debounce logic to user search

Pikalot:
Merge #1748 first.
This PR builds on top of PR2 (search UI) and introduces debounce logic to reduce API calls during user input.
Officer and Admin can search for users.
UI screenshots:
- Light mode:
image

- Dark mode:
image

### DIFF
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -552,8 +552,8 @@ router.post('/shortcutsearchusers', async function(req, res) {
   const tokenScores = (str, tokens) => {
     return tokens.reduce((score, token) => {
       if (str.startsWith(token)) return score + 0; // highest score for exact match
-      else if (str.includes(token)) return score + 1; // lower score for partial match
-      else return score + 2; // lowest score for no match
+      if (str.includes(token)) return score + 1; // lower score for partial match
+      return score + 2; // lowest score for no match
     }, 0);
   };
 
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -511,5 +511,96 @@ router.get('/getNewPaidMembersThisSemester', async (req, res) => {
   }
 });
 
+// Search for all members using either first name, last name or email
+router.post('/shortcutsearchusers', async function(req, res) {
+  if (!checkIfTokenSent(req)) {
+    return res.sendStatus(FORBIDDEN);
+  } else if (!checkIfTokenValid(req, membershipState.OFFICER)) {
+    return res.sendStatus(UNAUTHORIZED);
+  }
+
+  if (!req.body.query) {
+    return res.status(OK).send({ items: [] });
+  }
+
+  const query = req.body.query.replace(/[*\s]/g, '');
+
+  // Create a fuzzy regex pattern to match characters in order, e.g., "pone" -> /p.*o.*n.*e/i
+  const fuzzyPattern = query.split('').join('.*');
+  const pattern = new RegExp(fuzzyPattern, 'i');
+
+  const maybeOr = {
+    $or: [
+      {
+        $expr: {
+          $regexMatch: {
+            input: { $concat: ['$firstName', '$lastName'] },
+            regex: pattern,
+          }
+        }
+      },
+      { email: { $regex: new RegExp(query, 'i')} }
+    ]
+  };
+
+  /**
+   * Function to calculate scores based on token matches for sorting
+   * @param {string} str - The string to score against
+   * @param {Array} tokens - The tokens to match against the string
+   * @return {number} - The score based on matches
+   */
+  const tokenScores = (str, tokens) => {
+    return tokens.reduce((score, token) => {
+      if (str.startsWith(token)) return score + 0; // highest score for exact match
+      else if (str.includes(token)) return score + 1; // lower score for partial match
+      else return score + 2; // lowest score for no match
+    }, 0);
+  };
+
+  /**
+   * Sorts the user items based on the query match
+   * @param {string} query input string to match against
+   * @returns {function} - A comparison function for sorting
+   */
+  const sortByMatch = (query) => {
+    const input = query.toLowerCase().split(/[\s@._-]+/).filter(Boolean);
+
+    return (a, b) => {
+      const aName = (a.firstName + ' ' + a.lastName).toLowerCase();
+      const bName = (b.firstName + ' ' + b.lastName).toLowerCase();
+      const aEmail = a.email.toLowerCase();
+      const bEmail = b.email.toLowerCase();
+
+      // First Priority: sort by name match
+      const nameScoreA = tokenScores(aName, input);
+      const nameScoreB = tokenScores(bName, input);
+      if (nameScoreA !== nameScoreB) {
+        return nameScoreA - nameScoreB;
+      }
+
+      // Second Priority: sort by email match
+      const emailScoreA = tokenScores(aEmail, input);
+      const emailScoreB = tokenScores(bEmail, input);
+      if (emailScoreA !== emailScoreB) {
+        return emailScoreA - emailScoreB;
+      }
+
+      // Tie-breaker: alphabetical email sort
+      return a.email.localeCompare(b.email);
+    };
+  };
+
+  // Find user and sort results based on best match of full name or email
+  User.find(maybeOr, { password: 0 })
+    .limit(5)
+    .then(items => {
+      items.sort(sortByMatch(req.body.query));
+      res.status(OK).send({ items });
+    })
+    .catch((error) => {
+      logger.error('/shortcutsearchusers encountered an error:', error);
+      res.sendStatus(BAD_REQUEST);
+    });
+});
 
 module.exports = router;
--- a/src/Components/ShortcutKeyModal/SearchModal.js
+++ b/src/Components/ShortcutKeyModal/SearchModal.js
@@ -4,8 +4,9 @@ import { officerOrAdminRoutes, signedOutRoutes, memberRoutes, notAuthenticatedRo
 import { membershipState } from '../../Enums';
 import { useUser } from '../context/UserContext';
 import { useAuth } from '../context/AuthContext';
+import { searchAllUsers } from '../../APIFunctions/UserSearch';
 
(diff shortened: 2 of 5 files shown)


weslayer (approved):
boing boing

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+  // and avoid negative page numbers
+  let skip = Math.max(Number(req.body.page) || 0, 0);
+  skip *= ROWS_PER_PAGE;
+  const total = await User.count(maybeOr);
+  User.find(maybeOr, { password: 0, }, { skip, limit })
+    .sort({ [sortColumn] : sortOrder })
+    .then(items => {
+      res.status(OK).send({ items, total, rowsPerPage: ROWS_PER_PAGE, });
+    })
+    .catch((e) => {
```
adarshm11: you can remove the "e" in the catch, since you're not using it

src/Components/ShortcutKeyModal/SearchModal.js
```diff
@@ -4,6 +4,7 @@ import { officerOrAdminRoutes, signedOutRoutes, memberRoutes, notAuthenticatedRo
 import { membershipState } from '../../Enums';
 import { useUser } from '../context/UserContext';
 import { useAuth } from '../context/AuthContext';
+import { searchAllUsers } from '../../APIFunctions/UserSearch';
 
 export default function SearchModal({ appProps }) {
```
adarshm11: looks like appProps isn't used, is that by intention? if so then you can remove it

adarshm11 (approved):

src/Components/ShortcutKeyModal/SearchModal.js
Pikalot: Done! Removed

api/main_endpoints/routes/UserSearch.js
Pikalot: Done! Removed!

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  const limit = Math.min(Number(req.body.limit) || ROWS_PER_PAGE, ROWS_PER_PAGE);
+  let maybeOr = {};
+  const query = req.body.query;
+  if (query) {
+    const parts = query.trim().split(/\s+/);
+    const first = parts[0];
+    const last = parts[parts.length - 1];
+    const nameRegex = new RegExp(`^${first}.*${last}`, 'i');
```
evanugarte: what if we do a match using whatever the user typed

so if i was searching for `person one`, i could match it by typing `pone`

```suggestion
    const query = req.body.query.replace(/\s/g, '');
    const pattern = new RegExp(`.*${query}.*`);
```

another example the above would match on would be `person o`, as right now the pr's logic is expecting the entire first and last name to by typed out to do a match

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+
+    maybeOr = {
+      $or: [
+        {
+          $expr: {
+            $regexMatch: {
+              input: { $concat: ['$firstName', ' ', '$lastName'] },
```
evanugarte: see other comment about regex change, we can just do
```suggestion
              input: { $concat: ['$firstName', '$lastName'] },
```

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+  if (!checkIfTokenSent(req)) {
+    return res.sendStatus(FORBIDDEN);
+  } else if (!checkIfTokenValid(req, membershipState.OFFICER)) {
+    return res.sendStatus(UNAUTHORIZED);
+  }
+
+  const limit = Math.min(Number(req.body.limit) || ROWS_PER_PAGE, ROWS_PER_PAGE);
+  let maybeOr = {};
+  const query = req.body.query;
+  if (query) {
```
evanugarte: if the user didnt pass a query in should we even bother displaying suggestions?


```js
if (!req.body.query) {
  return res.status(OK).send({ items: [] });
}
```

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+          }
+        },
+        { firstName: { $regex: new RegExp(query, 'i')} },
+        { lastName: { $regex: new RegExp(query, 'i')} },
+        { email: { $regex: new RegExp(query, 'i')} }
+      ]
+    };
+  }
+
+  const sortColumn = req.query.sort || 'joinDate';
```
evanugarte: is there a way to have mongodb return the result based on best match? https://stackoverflow.com/a/27224578

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+
+  // make sure that the page we want to see is 0 by default
+  // and avoid negative page numbers
+  let skip = Math.max(Number(req.body.page) || 0, 0);
+  skip *= ROWS_PER_PAGE;
+  const total = await User.count(maybeOr);
+  User.find(maybeOr, { password: 0, }, { skip, limit })
+    .sort({ [sortColumn] : sortOrder })
+    .then(items => {
+      res.status(OK).send({ items, total, rowsPerPage: ROWS_PER_PAGE, });
```
evanugarte: total and rowsPerPage are unused in the frontend, can we remove it

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+    })
+    .catch(() => {
+      res.sendStatus(BAD_REQUEST);
```
evanugarte: import logger and log the error, would be good for debugging

api/main_endpoints/routes/UserSearch.js
```diff
@@ -0,0 +1,78 @@
...
+  BAD_REQUEST,
+  UNAUTHORIZED,
+  FORBIDDEN,
+} = require('../../util/constants').STATUS_CODES;
+const membershipState = require('../../util/constants').MEMBERSHIP_STATE;
+
+const ROWS_PER_PAGE = 20;
+
+// Search for all members using either first name, last name or email
+router.post('/shortcutsearchusers', async function(req, res) {
```
evanugarte: this api can go in `User.js` instead of a new file

api/main_endpoints/routes/UserSearch.js
Pikalot: New suggestion: Instead of matching the first and last name, the backend will match the record using the input sequence:
```
const query = req.body.query.replace(/\s/g, '');
const fuzzyPattern = query.split('').join('.*');
const pattern = new RegExp(fuzzyPattern, 'i');
```
This trades off O(n) performance for maximum flexibility. Note: this approach does not ignore discrepancies in middle names.

api/main_endpoints/routes/UserSearch.js
Pikalot: Done!

api/main_endpoints/routes/UserSearch.js
Pikalot: Done!

api/main_endpoints/routes/UserSearch.js
Pikalot: Done!

api/main_endpoints/routes/UserSearch.js
Pikalot: Done!

Implement scoring approach:
```
        const aScore = Math.min(tokenScores(aName, input), tokenScores(aEmail, input));
        const bScore = Math.min(tokenScores(bName, input), tokenScores(bEmail, input));

        return aScore - bScore; 
```

api/main_endpoints/routes/UserSearch.js
Pikalot: Done!

api/main_endpoints/routes/UserSearch.js
Pikalot: Done! Migrated codes to user.js

test/api/User.js
```diff
@@ -716,4 +716,204 @@ describe('User', () => {
...
+
+    it('Should return status code 403 if no token is passed through', async () => {
+      setTokenStatus(false);
+      const result = await test.sendPostRequest(url, queryUser);
+      expect(result).to.have.status(FORBIDDEN);
+    });
+
+    it('Should return status code 401 if access level is invalid', async () => {
+      accessLevel = MEMBERSHIP_STATE.MEMBER;
+      setTokenStatus(accessLevel >= requiredLevel, {accessLevel});
```
evanugarte: each line with `accessLevel >= requiredLevel`, please hardcode it to true/false, it will read much easier

api/main_endpoints/routes/User.js
```diff
@@ -511,5 +511,96 @@ router.get('/getNewPaidMembersThisSemester', async (req, res) => {
...
+   * Function to calculate scores based on token matches for sorting
+   * @param {string} str - The string to score against
+   * @param {Array} tokens - The tokens to match against the string
+   * @return {number} - The score based on matches
+   */
+  const tokenScores = (str, tokens) => {
+    return tokens.reduce((score, token) => {
+      if (str.startsWith(token)) return score + 0; // highest score for exact match
+      else if (str.includes(token)) return score + 1; // lower score for partial match
+      else return score + 2; // lowest score for no match
```
evanugarte:
```suggestion
      if (str.startsWith(token)) return score + 0; // highest score for exact match
      if (str.includes(token)) return score + 1; // lower score for partial match
      return score + 2; // lowest score for no match
```

evanugarte (approved):

api/main_endpoints/routes/User.js
Pikalot: Done!

test/api/User.js
Pikalot: Done!

adarshm11 (approved):

akshtshrma24 (approved):

PR #1752, opened 2025-07-04
remove user prop from index + urlshortener page

altran03:
1 prop down

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -29,7 +29,7 @@ function App(props) {
     !isAuthenticating && (
       <UserContext.Provider value={{ user, setUser }}>
         <BrowserRouter>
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -29,7 +29,7 @@ function App(props) {
     !isAuthenticating && (
       <UserContext.Provider value={{ user, setUser }}>
         <BrowserRouter>
-          <Routing appProps={{ authenticated, setAuthenticated, user }} />
+          <Routing appProps={{ authenticated, setAuthenticated }} />
```
evanugarte, line 34: we are sure this is the last page that needs the prop removed?

if so, this change is ok

evanugarte (approved):

weslayer (approved):
yay

PR #1755, opened 2025-07-05
Audit logs frontend

pxtrickhoxng:
Add audit logs UI to the admin dashboard, allowing officers and admins to view basic website info such as logins, sign-ups, user updates, and more. Sensitive info such as passwords are not logged.


adarshm11 (commented):
double check with evan about creating /Components and /utils folders inside src/Pages if that's the structure he's okay with. other than that it looks fine

PR #1760, opened 2025-07-08
Convert Advertisement expiration to a date input

### LINKED ISSUE #1727: Convert Advertisement expiration to a date input
the advertisement ui looks like:

Image

instead of 3 text boxes, lets change the input to be a single date input like below:

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/datetime-local

we can test this by making sure the advertisement is submitted with this new date argument, opening the network tab in the browser when submitting the url and checking the request body of the submit request

leishaaf:
resolves #1727

image

### DIFF
--- a/src/Pages/Advertisement/AdvertisementAdmin.js
+++ b/src/Pages/Advertisement/AdvertisementAdmin.js
@@ -46,35 +46,35 @@ export default function AdvertisementAdmin() {
   function maybeRenderExpirationInput(){
     if(!expireButtonClicked){
       return( <button
-      className="text-sm btn btn-primary sm:text-base"
-      onClick={() => setExpiredButtonClicked(true)}
-    >
+        className="text-sm btn btn-primary sm:text-base"
+        onClick={() => setExpiredButtonClicked(true)}
+      >
       Set Expiration Date For Ad
-    </button>
+      </button>
       );
     }else{
       return(
         <>
-      <input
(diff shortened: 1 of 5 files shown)


src/Pages/Advertisement/AdvertisementAdmin.js
```diff
@@ -101,30 +80,35 @@ export default function AdvertisementAdmin() {
...
-          />
-          <input
-            className="flex-1 text-sm input input-bordered sm:text-base"
-            type="text"
-            placeholder="Day"
-            onChange={event => {
-              setDay(event.target.value);
-            }}
-          />
+          {expireButtonClicked ? (
```
evanugarte: make this a function called `maybeRenderExpirationInput`

you can call it like
```
<div ...>
  {maybeRenderExpirationInput()}
</div>
```

and the function can have an in statement like

```js
if (!expireButtonClicked) {
  return <button ...> {/* button stuff here */} </button>
}
// down here return the input 
```

leishaaf: done!

evanugarte (approved):

thebeninator (approved):
lgtm

Embotic-Wayne (approved):
Amazing

armanbance (approved):
looks good!

PR #1769, opened 2025-07-09
use date input type

DanielTQuach:
#1727
use date input instead of text input


evanugarte:
we used this code in https://github.com/SCE-Development/Clark/pull/1760

Issue #1556, opened 2025-01-12
Allow users to specify a background color


evanugarte:
as part of the profile page, lets add a simple input that once edited and saved, the users background icon in the top right corner changes color.

### game plan
- [ ] follow dessert problem
- [ ] add a new field to the user.js model called `backgroundColor`
- [ ] add allow the user edit api to update the backgroundColor, if specified
- [ ] in the frontend, render the updated user background color, after the color is updated
- [ ] add logic in the profile icon to default to the current gray color if the users profile color is not specified
- [ ] add logic in the profile icon to determine if the text should be rendered white or black https://stackoverflow.com/a/3943023

PR #1774, opened 2025-07-13
Profile Icon Background Color

Funsize1234:
#1556 added ability for users to change their icon's background color from the profile page

ezgif-589d1622c73611

### DIFF
--- a/src/Pages/Profile/MemberView/Profile.js
+++ b/src/Pages/Profile/MemberView/Profile.js
@@ -1,25 +1,54 @@
 import React, { useEffect, useState } from 'react';
-import { getUserById } from '../../../APIFunctions/User';
+import { getUserById, editUser } from '../../../APIFunctions/User';
 import ChangePasswordModal from './ChangePassword';
 import DeleteAccountModal from './DeleteAccountModal';
 import GetApiKeyModal from './GetApiKeyModal';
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -31,10 +32,12 @@ function App() {
   return (
     !isAuthenticating && (
       <SceContext.Provider value={{user, setUser, authenticated, setAuthenticated}}>
-        <BrowserRouter>
-          <SearchModal/>
-          <Routing/>
-        </BrowserRouter>
+        <BackgroundColorContextProvider>
```
evanugarte, line 35: add it to SceContext.Provider

src/Pages/Profile/MemberView/Profile.js
```diff
@@ -1,25 +1,54 @@
...
 import ChangePasswordModal from './ChangePassword';
 import DeleteAccountModal from './DeleteAccountModal';
 import GetApiKeyModal from './GetApiKeyModal';
 import { membershipState, membershipStateToString } from '../../../Enums';
 import { useSCE } from '../../../Components/context/SceContext';
+import { useBackgroundColor } from '../../../Components/context/BackgroundColorContext';
 
 export default function Profile() {
+  const defaultColor = '#2a323c';
   const { user } = useSCE();
```
evanugarte, line 12: then you can use the context here

PR #1779, opened 2025-07-14
club revenue api function to use fetch instead of axios

Funsize1234:
image

this method is called on the overview page and yields the correct values in the response data.

### DIFF
--- a/src/APIFunctions/User.js
+++ b/src/APIFunctions/User.js
@@ -325,9 +325,7 @@ export async function getNewPaidMembersThisSemester(token) {
   const url = new URL('/api/User/getNewPaidMembersThisSemester', BASE_API_URL);
   try {
     const res = await fetch(url.href, {
-      method: 'GET',
       headers: {
-        'Content-Type': 'application/json',
         Authorization: `Bearer ${token}`,
       },
     });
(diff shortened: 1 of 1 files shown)


joshua-demo (approved):
lgtm

src/APIFunctions/User.js
```diff
@@ -322,14 +322,21 @@ export async function getApiKey(token) {
 
 export async function getNewPaidMembersThisSemester(token) {
   let status = new UserApiResponse();
+  const url = new URL('/api/User/getNewPaidMembersThisSemester', BASE_API_URL);
   try {
-    const url = new URL('/api/User/getNewPaidMembersThisSemester', BASE_API_URL);
-    const response = await axios.get(url.href, {
+    const res = await fetch(url.href, {
+      method: 'GET',
```
adarshm11: default method is get so you don't need to clarify this

src/APIFunctions/User.js
```diff
@@ -322,14 +322,21 @@ export async function getApiKey(token) {
...
       headers: {
-        Authorization: `Bearer ${token}`
-      }
+        'Content-Type': 'application/json',
```
adarshm11: this header needed for post request, not get, so you can remove it

adarshm11 (approved):
good work

src/APIFunctions/User.js
Funsize1234: done

src/APIFunctions/User.js
Funsize1234: done

weslayer (approved):

PR #1790, opened 2025-07-15
pass print job id to client

thebeninator:
quasar returns a json obj w/ print id that needs to be passed to the client

### DIFF
--- a/api/main_endpoints/routes/Printer.js
+++ b/api/main_endpoints/routes/Printer.js
@@ -84,7 +84,7 @@ router.post('/sendPrintRequest', upload.single('chunk'), async (req, res) => {
   }
   if (!PRINTING.ENABLED) {
     logger.warn('Printing is disabled, returning 200 and dummy print id to mock the printing server');
-    return res.status(OK).send({ print_id: null });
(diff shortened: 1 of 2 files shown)


api/main_endpoints/routes/Printer.js
```diff
@@ -122,16 +122,17 @@ router.post('/sendPrintRequest', upload.single('chunk'), async (req, res) => {
...
     // full pdf can be sent to quasar no problem
-    await axios.post(PRINTER_URL + '/print', data, {
+    const printRes = await axios.post(PRINTER_URL + '/print', data, {
       headers: {
         ...data.getHeaders(),
       },
       maxContentLength: 1024 * 1024 * 150, // 150 mb
       maxBodyLength: Infinity
     });
+    const printId = printRes.data;
```
evanugarte, line 134: add a comment above for what data looks like, i.e. what the json object returned from quasar looks like

evanugarte (approved):

api/main_endpoints/routes/Printer.js
thebeninator: done

weslayer (approved):

PR #1791, opened 2025-07-15
add basic audit logs page for admin dashboard (JSON objects, no formatting)

pxtrickhoxng:
image

image

This is the first of three PRs for the audit logs page.

PR #1792, opened 2025-07-15
Audit logs frontend formatting

pxtrickhoxng:
image

#1791 should be merged first.


pxtrickhoxng:
I noticed just now that utils/activityTypes.js contains a few extra activities that the audit log schema doesn't contain. These are removed on PR #1793, but they shouldn't cause any problems right now; they'll just be unused.

src/Pages/AuditLog/Components/AuditLogCard.js
```diff
@@ -0,0 +1,115 @@
+const AuditLogCard = ({ log, index }) => {
+  const hasDetails = log => {
+    return log.details && Object.keys(log.details).length > 0;
+  };
+
+  const formatDetails = details => {
+    if (!details || Object.keys(details).length === 0) {
+      return null;
+    } else {
```
evanugarte: no need for the else block since we return in the if statement

evanugarte (approved):

adarshm11 (approved):
lgtm

PR #1802, opened 2025-07-20
LED sign page, fixed text being white on light mode

Matreeso:
Previously, when using light mode for the LED sign page, the user inputted text would be white when using light mode, making it unreadable.

Now that issue has been fixed, and the text is now readable in both light mode and dark mode.


evanugarte (approved):

adarshm11 (approved):
big if true

PR #1814, opened 2025-07-24
implemented mongoose-fuzzy-searching library

aspies06:


evanugarte:
we are gonna reuse these changes in a different pr, the work and effort is not lost

PR #1820, opened 2025-07-24
Replace redis w/mongodb

### LINKED ISSUE #1777: Replace redis with Mongodb TTL feature for password reset
- [ ] work with @durpdur on setting up gmail, see https://github.com/SCE-Development/Clark/wiki/Getting-Started#setting-up-gmail-skip-if-you-are-a-new-dev-member
- [ ] make a mongodb schema with two fields, resetToken and userId, resetToken's definition should include `index: true,` for fast querying.
- [ ] add a TTL index to your schema, so the entry in the colleciton expires after a specified time, like https://github.com/SCE-Development/Clark/blob/9223638b6a31532b935721d167456e3308855f94/api/main_endpoints/models/Advertisement.js#L15
- [ ] change all uses of redisClient in https://github.com/SCE-Development/Clark/blob/3cc142973c3f6fc78ad53a63894a3b383625f4a9/api/main_endpoints/routes/Auth.js#L119 to be this new schema, where we create and delete fields. be sure to also set the TTL index.
- [ ] change all uses of redisClient in https://github.com/SCE-Development/Clark/blob/3cc142973c3f6fc78ad53a63894a3b383625f4a9/api/main_endpoints/util/userHelpers.js#L173 to this new schema usage
- [ ] uninstall redis from clark, open pr.

altran03:
Replacing redis with mongodb for password reset


weslayer (approved):
is this the only usage of redis in the project

docker-compose.yml
```diff
@@ -129,4 +122,3 @@ networks:
       name: sce
 
 volumes:
```
evanugarte:
```suggestion
```

evanugarte (approved):

PR #1822, opened 2025-07-24
Added ChatMessage schema & fixed api logic so it actually stores each chatMessage in the db

Sujan30:
- Created Chat Message schema, has createdAt, expiresAt, text, userId, chatroomId
- added logic to make sure that each chatMessage expires exactly after 24h- > default: ()=> Date.now() + 24 * 3600 * 1000, 
- fixed write message method in api/main_endpoints/routing/Message.js to actually write each chatmessage into the db using the above schema
- fixed getLatestMessage, so for each chatroom id, we can get the most recent messages.
- Used users username to query the User db to get the userid which is then used to be a part of the ChatMessage db

Screenshot 2025-07-30 at 2 45 51 AM

This is me typing messages into the chatroom with id= hello

here is the /getLatestMessage endpoint showing the latest messages
Screenshot 2025-07-30 at 2 44 55 AM

here is /getLatestMessage with TTL set to 30s

30 seconds later...

Screenshot 2025-07-30 at 2 45 42 AM

### DIFF
--- a/api/main_endpoints/models/ChatMessage.js
+++ b/api/main_endpoints/models/ChatMessage.js
@@ -1,0 +1,36 @@
+const mongoose = require('mongoose')
+const Schema = mongoose.Schema;
+
+
+const ChatMessageSchema = new Schema(
+    {
+        createdAt: {
+            type: Date,
+            default: Date.now(),
+        }, 
+        expiresAt: {
+            type: Date, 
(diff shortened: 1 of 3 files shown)


weslayer (approved):

charred-70 (commented):
ltgm so coolio

charred-70 (approved):
Oh wait here's the actual review lol

PR #1836, opened 2025-07-28
Feature/secure print

Embotic-Wayne:


api/main_endpoints/util/userHelpers.js
```diff
@@ -202,6 +202,23 @@ function checkIfPageCountResets(lastLogin) {
...
+  const user = await User.findById(userId);
+  if(!user){
+    throw new Error('user not found'); // checks user
+  }
+  if(!Number.isInteger(pagesToPrint) || pagesToPrint <= 0){
+    throw new Error('invalid number');
+  }
+  if (user.pagesPrinted < pagesToPrint) {
+    throw new Error('no pages remaining');
+  }
```
evanugarte: instead of throwing an error just do logger.error here and return early, we dont have to return anything

api/main_endpoints/routes/Printer.js
```diff
@@ -114,7 +117,18 @@ router.post('/sendPrintRequest', upload.single('chunk'), async (req, res) => {
...
   }
 
-  const stream = await fs.createReadStream(assembledPdfFromChunks);
+  try{
+    const stream = await fs.promises.readFile(assembledPdfFromChunks); // buffer
+    const pdfDoc = await PDFDocument.load(stream); // load PDF
+    const numpages = pdfDoc.getPages().length; // get number of pages
+    const copiesInt = parseInt(copies || 1, 10);
+    const totalPages = numpages * copiesInt;
+    await subtractUserPages(user.id, totalPages); // updates users printcount
```
evanugarte: move this code in the existing try block in `axios.post(PRINTER_URL ...`

efar301 (approved):
very good

PR #1838, opened 2025-07-28
fuzzy search

andresaclan:


andresaclan:
Implemented some of @Pikalot 's changes. Updated test cases to pass with new fuzzy search.

PR #1861, opened 2025-07-31
Implemented levenshtein distance algorithm and changed unit tests

aspies06:


aspies06:
Needs #1850 to be merged first in order for api test to work properly

PR #1864, opened 2025-07-31
Added api/AuditLog.js and a test case to frontend.test.js

Pikalot:
frontend-test:
image

===================================================================================

api-test:
image

### DIFF
--- a/test/api/AuditLog.js
+++ b/test/api/AuditLog.js
@@ -1,0 +1,170 @@
+process.env.NODE_ENV = 'test';
+
+const User = require('../../api/main_endpoints/models/User.js');
+const AuditLog = require('../../api/main_endpoints/models/AuditLog.js');
+
+const chai = require('chai');
+
+const chaiHttp = require('chai-http');
+const sinon = require('sinon');
+const {
+  OK,
+  BAD_REQUEST,
+  UNAUTHORIZED,
+  FORBIDDEN,
+  NOT_FOUND,
+  CONFLICT,
+  SERVER_ERROR,
+} = require('../../api/util/constants.js').STATUS_CODES;
+const SceApiTester = require('../util/tools/SceApiTester.js');
+
+let app = null;
+let test = null;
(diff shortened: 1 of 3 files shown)


test/api/AuditLog.js
```diff
@@ -0,0 +1,170 @@
...
+    it('Should return status code 401 if access level is invalid', async () => {
+      setTokenStatus(false, { accessLevel: MEMBERSHIP_STATE.MEMBER });
+      const result = await test.sendGetRequestWithToken(token, url);
+      expect(result).to.have.status(UNAUTHORIZED);
+    });
+
+    describe('When token and access level is valid - status code 200', () => {
+      beforeEach(() => {
+        setTokenStatus(true, { accessLevel: MEMBERSHIP_STATE.OFFICER });
+      });
```
adarshm11: do we need a beforeEach inside the test

test/api/AuditLog.js
```diff
@@ -0,0 +1,170 @@
...
+
+        for (let i = 0; i < 60; i++) {
+          await AuditLog.create({
+            userId: newUser._id,
+            action: AuditLogActions.EMAIL_SENT,
+            documentId: newUser._id,
+            details: {email: newUser.email },
+          });
+        }
+      });
```
adarshm11, line 130: what are these loops for?

Pikalot: The loops create 63 records. The test checks that, when the query is empty, the route returns at most 50 records from the first page.

test/api/AuditLog.js
Pikalot: I'm afraid it won't work without this block, since the beforeEach in the outer loop overrides all token setup. If this approach slows down the performance badly, we can consider restructuring the test.

Pikalot: Done! Removed the beforeEach() block and moved all 200-status-code test cases into the same describe block with other test cases.

PR #1869, opened 2025-08-01
Added create audit log test in printer.js

Pikalot:
This PR is in sync with PR # https://github.com/pxtrickhoxng/Clark/commits/audit-on-print/

Screenshot:
image


api/main_endpoints/routes/Printer.js
```diff
@@ -74,29 +78,42 @@ router.get('/healthCheck', async (req, res) => {
...
-  if (!PRINTING.ENABLED) {
-    logger.warn('Printing is disabled, returning 200 and dummy print id to mock the printing server');
-    return res.status(OK).send({ printId: null });
-  }
 
-  const dir = path.join(__dirname, 'printing');
-  const { totalChunks, chunkIdx } = req.body;
+  const details = {
+    copies: parseInt(copies),
+    sides: sides,
```
adarshm11:
```suggestion
    sides,
```
when key = value we only need to write it once

api/main_endpoints/routes/Printer.js
```diff
@@ -74,29 +78,42 @@ router.get('/healthCheck', async (req, res) => {
 });
 
 router.post('/sendPrintRequest', upload.single('chunk'), async (req, res) => {
+  let totalFileSize = 0;
+  const { copies, sides, id } = req.body;
+  const action = AuditLogActions.PRINT_PAGE;
```
adarshm11: let's just remove this and specify what the AuditLog action is when we create the AuditLog

api/main_endpoints/routes/Printer.js
```diff
@@ -74,29 +78,42 @@ router.get('/healthCheck', async (req, res) => {
...
-  if (Number(chunkIdx) < totalChunks - 1) {
+  if (!PRINTING.ENABLED) {
+    details.status = 'mocked';
+    // create audit log on print
+    await createAuditLog({
+      user,
+      action,
+      details
+    });
+    logger.warn('Printing is disabled, returning 200 to mock the printing server');
```
adarshm11: is this creating an audit log when we're running Clark on local? if so, i don't think we need this, we can find a way to write unit tests around this specific case

api/main_endpoints/util/auditLogHelpers.js
```diff
@@ -0,0 +1,38 @@
...
+    logger.error('auditLogHelpers had an error', err);
+    if (err.response && err.response.data) {
+      return {
+        status: err.response.status,
+        error: err.response.data
+      };
+    } else {
+      return {
+        status: SERVER_ERROR,
+        error: 'Failed to create audit log in cleezyHelpers'
```
adarshm11: are we only using this for cleezy? if not then we should change this line

api/main_endpoints/util/auditLogHelpers.js
```diff
@@ -0,0 +1,38 @@
...
+      };
+    }
+  }
+
+};
+module.exports = { createAuditLog };
```
adarshm11: if we are creating a helper function to create Audit logs, we need to notify the Audit Log team so that this can be used everywhere Audit Logs are being created across Clark

api/main_endpoints/routes/Printer.js
Pikalot: Done!

api/main_endpoints/util/auditLogHelpers.js
Pikalot: Nice catch! My poor copy and paste mistakes! 
Done! Removed the helper for now and waiting for the Audit team's idea!

api/main_endpoints/util/auditLogHelpers.js
Pikalot: Done! Removed the helper for now and waiting for the Audit team's idea!

api/main_endpoints/routes/Printer.js
Pikalot: Done!

api/main_endpoints/routes/Printer.js
Pikalot: Done! This block is removed.

PR #1876, opened 2025-08-05
Added prometheus metric 

durpdur:
Added Prometheus metric to  track current epoch time and when refresh token needs to be refreshed.


adarshm11 (approved):
fix da merge conflict before merge

evanugarte (approved):
image

PR #1877, opened 2025-08-08
Correct decodeToken behavior

adarshm11:
we don't really need any token functions besides `decodeToken` to check if 
- user is authenticated
- user's authentication level meets the minimum requirement for role-specific functions (e.g. only officers can delete office access cards)

so we can delete all other token functions and absorb them into `decodeToken`


evanugarte (approved):

weslayer (approved):
adersh

vietnguyen2358 (approved):
LGTM

thebeninator (approved):
good

PR #1884, opened 2025-08-15
fixed some misnamed html attributes

adarshm11:
couple misnamed html attributes that i caught so i fixed em


n8thantran (approved):
wow!

MatthewTran22 (approved):
LGTM 🤓

PR #1906, opened 2025-09-16
add metric to see which endpoint is not loaded 

luuisabelle:


evanugarte (approved):
its better than good

adarshm11 (approved):
yippee

PR #1928, opened 2025-10-13
remove unused alumni role

adarshm11:
nobody b using ts site with alumni role anyway


weslayer (approved):

n8thantran (approved):
this is the most amazing pr of all time u r the goat

PR #1960, opened 2025-12-16
Add live client-side search to user directory

jask1m:
In the admin dashboard, the search experience for user dashboard only filters candidates upon clicking `Enter`. This PR improves the search experience by filtering realtime on each query keystroke. This workaround also reduces the frequency of API calls to the initial load and delete requests rather than at each search filter (because filtering happens on client now).

* Implemented instant search that filters users as they type in the search box
* Fetch all users once on page load using page: -1 parameter instead of paginated requests
* Updated backend to support page: -1 for fetching all users without pagination

***Testing:***
Sign in with: (email: test@one.sce, pw: sce)
link: https://unstultifying-barrable-patty.ngrok-free.dev/

### DIFF
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -143,12 +143,24 @@ router.post('/users', async function(req, res) {
   };
   const sortOrder = orderToInteger[req.query.order] || orderToInteger.default;
 
-  // make sure that the page we want to see is 0 by default
-  // and avoid negative page numbers
-  let skip = Math.max(Number(req.body.page) || 0, 0);
-  skip *= ROWS_PER_PAGE;
+  // Handle pagination: defaults to page 0 if not specified
+  // Special case: page -1 fetches all users without pagination
+  // All other negative page numbers are clamped to 0
+  let skip, limit;
+  const pageNum = Number(req.body.page);
+
+  if (pageNum === -1) {
+    // Fetch all users (no pagination)
+    skip = 0;
+    limit = 0; // MongoDB uses 0 to mean "no limit"
+  } else {
+    // Regular pagination: clamp negative pages to 0
+    skip = Math.max(pageNum || 0, 0) * ROWS_PER_PAGE;
+    limit = ROWS_PER_PAGE;
+  }
+
   const total = await User.count(maybeOr);
-  User.find(maybeOr, { password: 0, }, { skip, limit: ROWS_PER_PAGE, })
+  User.find(maybeOr, { password: 0, }, { skip, limit })
     .sort({ [sortColumn] : sortOrder })
     .then(items => {
       res.status(OK).send({ items, total, rowsPerPage: ROWS_PER_PAGE, });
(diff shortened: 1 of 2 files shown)


src/Pages/Overview/Overview.js
```diff
@@ -89,25 +78,61 @@ export default function Overview() {
...
+    // Filter users based on query
+    let filtered = allUsers;
+    if (query.trim()) {
+      const searchTerm = query.trim().toLowerCase();
+      filtered = allUsers.filter(user => {
+        return (
+          user.firstName?.toLowerCase().includes(searchTerm) ||
+          user.lastName?.toLowerCase().includes(searchTerm) ||
+          user.email?.toLowerCase().includes(searchTerm)
+        );
```
evanugarte, line 99: why not have the backend do this filtering? thats what wee did before this pr

src/Pages/Overview/Overview.js
```diff
@@ -89,25 +78,61 @@ export default function Overview() {
...
+        } else {
+          // Handle dates
+          const dateA = new Date(aVal);
+          const dateB = new Date(bVal);
+          comparison = dateA.getTime() - dateB.getTime();
+        }
+
+        return currentSortOrder === 'asc' ? comparison : -comparison;
+      });
+    }
```
evanugarte, line 129: 100% the job of the backend, not here

api/main_endpoints/routes/User.js
```diff
@@ -143,12 +143,24 @@ router.post('/users', async function(req, res) {
...
+  // Special case: page -1 fetches all users without pagination
+  // All other negative page numbers are clamped to 0
+  let skip, limit;
+  const pageNum = Number(req.body.page);
+
+  if (pageNum === -1) {
+    // Fetch all users (no pagination)
+    skip = 0;
+    limit = 0; // MongoDB uses 0 to mean "no limit"
+  } else {
```
evanugarte, line 156: is there a situation where we want to fetch every user ever

the pagination approach we have seems fine to me

evanugarte (commented):
holy vibecoded 😭✌️

src/Pages/Overview/Overview.js
jask1m: Yes, but I filter here on client so I can do a live search result update as user edits query, rather than making an API after a user clicks enter each time.

api/main_endpoints/routes/User.js
jask1m: The rationale for this was that the initial load fetches all users and the client will handle filtering. I think the other approach to get the ux I want for search would be to debounce api calls as user types query but filtering on client seemed more appropriate.

evanugarte: imagine its 3000+ rows to sort on, every time the user changes their input, thats a lot to ask for the browser when the backend is attached to mongodb

can we move this to the backend, and keep the pagination approach, we only ever need to return like the first 25 records

PR #1972, opened 2025-12-27
Add PermissionRequest schema and API for LED sign access requests

Embotic-Wayne:
- added mongodb schema for user id, type, and when its deleted (excluding duplicates)
- new api to create, read, and delete permission requests sent by members
- unit tests to validate their behavior


charred-70 (approved):
lgtm w tree

adarshm11 (approved):
is that a banana

api/main_endpoints/models/PermissionRequest.js
```diff
@@ -0,0 +1,30 @@
...
+const Schema = mongoose.Schema;
+const PermissionRequestTypes = require('../util/permissionRequestTypes');
+
+const PermissionRequestSchema = new Schema(
+  {
+    userId: {
+      type: Schema.Types.ObjectId,
+      ref: 'User',
+      required: true,
+      index: true,
```
evanugarte: we can remove this, we have the `PermissionRequestSchema.index` to take care of this

api/main_endpoints/models/PermissionRequest.js
```diff
@@ -0,0 +1,30 @@
...
+    },
+    type: {
+      type: String,
+      enum: Object.keys(PermissionRequestTypes),
```
evanugarte: i know they will be a 1 to 1 mapping, we will use the values when we create new data though
```suggestion
      enum: Object.values(PermissionRequestTypes),
```

api/main_endpoints/models/PermissionRequest.js
```diff
@@ -0,0 +1,30 @@
...
+    deletedAt: {
+      type: Date,
+      default: null,
+    },
+  },
+  { timestamps: { createdAt: true, updatedAt: false } }
+);
+
+// Compound unique index prevents duplicate active requests per user+type
+PermissionRequestSchema.index({ userId: 1, type: 1 }, { unique: true });
```
evanugarte: instead of { unique: true } can it be
```js
{ 
    unique: true, 
    partialFilterExpression: { deletedAt: null } 
}
```

that way we dont index deleted records

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  try {
+    const permissionRequest = await PermissionRequest.create({
+      userId: decoded.token._id,
+      type,
+    });
+    const populated = await PermissionRequest.findById(permissionRequest._id)
+      .populate('userId', 'firstName lastName email');
+    res.status(OK).send(populated);
```
evanugarte: we can also return an empty 200, unless we want to do something with this data in the frontend after submitting

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,94 @@
...
+  } catch (error) {
+    if (error.code === 11000) return res.status(BAD_REQUEST).send({ error: 'Request already exists' });
```
evanugarte: we have `CONFLICT` instead of `BAD_REQUEST`

we can send that and no text back, since conflict implies something exists

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,94 @@
...
+
+    if (!request) return res.sendStatus(NOT_FOUND);
+    res.status(OK).send(request);
+  } catch (error) {
+    logger.error('Failed to get permission request:', error);
+    res.sendStatus(SERVER_ERROR);
+  }
+});
+
+router.post('/delete', async (req, res) => {
```
evanugarte, line 65: should admins be able to delete records, and whatever record they want?

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,94 @@
...
+      .populate('userId', 'firstName lastName email');
+    res.status(OK).send(populated);
+  } catch (error) {
+    if (error.code === 11000) return res.status(BAD_REQUEST).send({ error: 'Request already exists' });
+    logger.error('Failed to create permission request:', error);
+    res.sendStatus(SERVER_ERROR);
+  }
+});
+
+router.get('/getAll', async (req, res) => {
```
evanugarte: how about `/` returns everything, up to you

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,94 @@
...
+      .populate('userId', 'firstName lastName email')
+      .sort({ createdAt: -1 });
+    res.status(OK).send(requests);
+  } catch (error) {
+    logger.error('Failed to get permission requests:', error);
+    res.sendStatus(SERVER_ERROR);
+  }
+});
+
+router.get('/get', async (req, res) => {
```
evanugarte: can we consolidate both endpoints into one `/`

the endpoint can either take a parameter like userId, or nothing

if we didnt pass any filters in, return everything if the user is an officer or higher

if we passed in a user id, return the data if
- the user is less than an officer and its their own user id.
- the user is an officer or higher

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,109 @@
...
+      if (!isOfficer) {
+        return res.sendStatus(UNAUTHORIZED);
+      }
+    } else {
+      // If there is a userId, check their perms
+      if (!isOfficer && queryUserId !== decoded.token._id.toString()) {
+        return res.sendStatus(FORBIDDEN);
+      }
+      query.userId = queryUserId;
+    }
```
evanugarte: sorry how about this

if the user is not an officer, dont listen to any id they send them, and just use `decoded.token._id.toString()` as what to query for

evanugarte (approved):

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+});
+
+router.post('/delete', async (req, res) => {
+  const decoded = await decodeToken(req, membershipState.MEMBER);
+  if (decoded.status !== OK) return res.sendStatus(decoded.status);
+
+  const { type, _id } = req.body;
+  if (!type || !Object.keys(PermissionRequestTypes).includes(type)) {
+    return res.status(BAD_REQUEST).send({ error: 'Invalid type' });
```
evanugarte:
```suggestion
    return res.status(BAD_REQUEST).send({ error: `${type} is an invalid type, try ${Object.keys(PermissionRequestTypes)}` });
```

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,104 @@
...
+        deletedAt: null,
+      });
+    } else {
+      // Members can delete their own requests and officers can delete their own requests without id
+      request = await PermissionRequest.findOne({
+        userId: decoded.token._id,
+        type,
+        deletedAt: null,
+      });
+    }
```
evanugarte: lets do this like below, also before this, lets make sure _id is set, and if it isnt, use the decoded.token._id

```js
let idToUse = _id;

if (decoded.token.accessLevel < membershipState.OFFICER) {
  idToUse = decoded.token._id;
}

const query = {
  _id: idToUse,
  type,
  deletedAt: null,
};

const request = await PermissionRequest.findOne({
        userId: decoded.token._id,
        type,
        deletedAt: null,
      });
```

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,104 @@
...
+
+    if (!request) return res.sendStatus(NOT_FOUND);
```
evanugarte, line 94: are we sure if no record if found, it returns null? lets add a unit test to /delete to ensure 404 is returned when we search for an _id that doesnt exist

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,104 @@
...
+    });
+    res.sendStatus(OK);
+  } catch (error) {
+    if (error.code === 11000) return res.sendStatus(CONFLICT);
+    logger.error('Failed to create permission request:', error);
+    res.sendStatus(SERVER_ERROR);
+  }
+});
+
+router.get('/get', async (req, res) => {
```
evanugarte:
```suggestion
router.get('/', async (req, res) => {
```

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -0,0 +1,104 @@
...
+  try {
+    const query = { deletedAt: null };
+
+    if (!isOfficer) {
+      query.userId = decoded.token._id.toString();
+    } else {
+      if (queryUserId) {
+        query.userId = queryUserId;
+      }
+    }
```
evanugarte, line 47: this should also work
```suggestion
    if (queryUserId) {
      query.userId = queryUserId;
    }
    if (!isOfficer) {
      query.userId = decoded.token._id.toString();
    }
```

evanugarte (approved):

PR #1976, opened 2025-12-29
add handleCancel to card reader confirmation modal

adarshm11:
before it used to bug out if u delete, then cancel, then delete again, now its fixed


charred-70 (approved):
Lgtm w adarshmellow

thebeninator (approved):
lgtm

PR #1981, opened 2025-12-31
leetcode leaderboard page

adarshm11:
## Summary
This PR includes a page and API for managing the LeetCode Leaderboard (sign2) from Clark.

## Features
- The leaderboard management page can be accessed from the admin navbar. Officers and admins can add/remove users, and view a preview of what the leaderboard currently looks like. 
- The API includes routes to add, delete, and fetch all users, as well as utility functions to perform these actions on the actual sign2 server. 
- Unit tests are included - however, these only test that the API routes behave correctly regarding JWT authentication, and does not test the routes' behavior with actually performing requests to sign2. This is because the sign2 connection is behind a gateway controlled by `config.json`, where users can set whether it is enabled or not. In the future, further unit tests should be added to address these conditions.

## Testing
This feature has been tested on the production server, where all functions work correctly. Ensure that, after merging, the sign2-specific data in `/api/config/config.json` matches what is expected to make sign2 reachable from Clark.


n8thantran (approved):
lgtm

api/config/config.example.json
```diff
@@ -32,6 +32,7 @@
   "LED_SIGN": {
     "ENABLED": false
   },
+  "LEETCODE_URL": "NOT_SET",
```
evanugarte: would this change in development vs prod

if it wouldnt, can we just do an ENABLED true/false

api/main_endpoints/routes/LeetCodeLeaderboard.js
```diff
@@ -0,0 +1,108 @@
...
+  getAllUsers,
+  addUserToLeaderboard,
+  deleteUserFromLeaderboard,
+  checkIfUserExists,
+} = require('../util/LeetCodeLeaderboard.js');
+const AuditLogActions = require('../util/auditLogActions.js');
+const AuditLog = require('../models/AuditLog.js');
+const membershipState = require('../../util/constants').MEMBERSHIP_STATE;
+
+router.get('/getAllUsers', async (req, res) => {
```
evanugarte: make life simple
```suggestion
router.get('/', async (req, res) => {
```

api/main_endpoints/routes/LeetCodeLeaderboard.js
```diff
@@ -0,0 +1,108 @@
...
+  const decoded = await decodeToken(req, membershipState.OFFICER);
+  if (decoded.status !== OK) {
+    return res.sendStatus(decoded.status);
+  }
+
+  const { username, firstName, lastName } = req.body;
+  const required = [
+    { value: username, title: 'LeetCode username', },
+    { value: firstName, title: 'User\'s first name', },
+    { value: lastName, title: 'User\'s last name', }
```
evanugarte, line 53: do we need first and last name? up to you

if we just do username does that hurt anything, would be simpler

src/Pages/LeetCodeLeaderboard/LeetCodeLeaderboard.js
```diff
@@ -0,0 +1,227 @@
...
+      setIsError(true);
+      setMessage('This username is already registered, please try again');
+      return;
+    }
+    const newUser = {
+      username: leetcodeUsername,
+      firstName,
+      lastName
+    };
+    if (!await addUser(newUser, token)) {
```
evanugarte: isntead of a second endpoint you could have the create api check if they exist

the create api can return 409, and if the frontend gets 409 then we just say they already are registered

api/config/config.example.json
adarshm11: 👍

api/main_endpoints/routes/LeetCodeLeaderboard.js
adarshm11: 👍

api/main_endpoints/routes/LeetCodeLeaderboard.js
adarshm11: names = easy to tell who's who on frontend
like who gonna know who "doggymuncher" is if there's no names you get what i mean

src/Pages/LeetCodeLeaderboard/LeetCodeLeaderboard.js
adarshm11: 👍

api/config/config.example.json
```diff
@@ -32,6 +32,9 @@
   "LED_SIGN": {
     "ENABLED": false
   },
+  "SIGN2": {
```
evanugarte:
```suggestion
  "LEETCODE_LED_SIGN": {
```

api/main_endpoints/routes/LeetCodeLeaderboard.js
```diff
@@ -0,0 +1,108 @@
+const express = require('express');
+const router = express.Router();
+const {
+  OK,
+  SERVER_ERROR,
+  BAD_REQUEST
+} = require('../../util/constants.js').STATUS_CODES;
+const { decodeToken } = require('../util/token-functions.js');
+const logger = require('../../util/logger.js');
```
evanugarte, line 9: can we just have this be a more dynamic file, basically anything that goes to 

/api/LeetCodeLeaderboard/...

routes whatever is at `...` to sign2

this file would then just
1. verify token
1. audit log based on route and request body
1. return whatever the response is to sign2

if the json is disabled, return 200, before doing any of the above steps

evanugarte (commented):
this pr could probably be way less code, see comment

api/main_endpoints/routes/LeetCodeLeaderboard.js
```diff
@@ -0,0 +1,120 @@
...
+const AuditLogActions = require('../util/auditLogActions.js');
+const AuditLog = require('../models/AuditLog.js');
+const membershipState = require('../../util/constants').MEMBERSHIP_STATE;
+const { LEETCODE_LED_SIGN = {} } = require('../../config/config.json');
+const axios = require('axios');
+const LEETCODE_LED_SIGN_URL = process.env.LEETCODE_LED_SIGN_URL || 'http://localhost:12121';
+
+router.get('/', async (req, res) => {
```
evanugarte: i think this file could be even less

like
```js
app.route('/:id')
  .get((req, res) => {
    const userId = req.params.id; // Access dynamic parameter
  })
  .post((req, res) => {
    const userId = req.params.id;
  });
```

you could probably make a single callback, and pass it in to both .get and .post

like

```js
function idk(req, res) {
  // ...
}

app.route('/users/:id')
  .get(idk)
  .post(idk);
```

change :id with :path or something

api/main_endpoints/routes/LeetCodeLeaderboard.js
```diff
@@ -0,0 +1,86 @@
...
+
+    return response.data;
+  } catch (err) {
+    logger.error(`Error with LeetCode Leaderboard at ${endpoint}: `, err.message || err);
+    res.status(SERVER_ERROR).send('Error communicating with LeetCode Leaderboard');
+    return null;
+  }
+};
+
+router.get('/', async (req, res) => {
```
evanugarte, line 51: do we need this code still, if we have the router.use

does router.use take care of all this stuff now

src/Components/Navbar/AdminNavbar.js
```diff
@@ -144,6 +144,15 @@ export default function UserNavBar(props) {
         </svg>
       )
     },
+    {
+      title: 'LeetCode Leaderboard',
+      route: '/leetcode-leaderboard',
```
evanugarte: this would be easier to type manually, i usually lookup stuff using url input in browser
```suggestion
      route: '/leetcode',
```

nginx.conf
```diff
@@ -149,6 +149,14 @@ http {
             return 404 "oops! we couldnt load your interview question lol, i guess this ones free!";
         }
 
+        location /sign2/ {
```
evanugarte, line 152: is this for seeing the leaderboard data directly on sce.sjsu.edu ?

like the emulator, you can resolve this comment if so

api/main_endpoints/routes/LeetCodeLeaderboard.js
adarshm11: image
we can do it like this but it doesn't seem aesthetically pleasing to me, thoughts

evanugarte: lets do it it seems clean

docker/Dockerfile
```diff
@@ -12,6 +12,7 @@ COPY ./package-lock.json /frontend/package-lock.json
 # and instead define them inline here. We do this as multi-stage
 # builds and environment variables from docker-compose.yml weren't working.
 ENV REACT_APP_BASE_API_URL https://sce.sjsu.edu
+ENV REACT_APP_SIGN2_URL /sign2/
```
evanugarte, line 15: we dont use this anywhere do we need it

PR #1982, opened 2026-01-01
add FORBIDDEN to token stub tests

adarshm11:
right now we can only stub tokens to be good (OK) or bad (UNAUTHORIZED), which ignores the possibility that a token is valid but possesses invalid access level (FORBIDDEN would be expected here). this PR allows for tokens to be mocked to return FORBIDDEN status


evanugarte (approved):

charred-70 (approved):
Waow waow W adarshmellow

altran03 (approved):
W Adarsh 😀

Issue #1952, opened 2025-12-15
[placeholder] dynamic, newer homepage image


evanugarte:
a random picture that appears on page load, can be
- 2025 internship picture
- room empty
- room full
- people decorating room
- office picture
- dev team collaboration

and more

maybe even a website counter to count number of people who visited

PR #1985, opened 2026-01-02
visit counter endpoint + model, display on homepage

jctimbol:
for issue #1952


adarshm11:
add a screenshot!!

jctimbol:
Screenshot 2026-01-03 at 9 27 42 PM

on bottom right of the home page

adarshm11:
looks good, make sure you fix lint errors, you can run `npm run lint` to see what's failing and where

api/main_endpoints/models/HomepageVisit.js
```diff
@@ -0,0 +1,14 @@
+const mongoose = require('mongoose');
+const Schema = mongoose.Schema;
+
+const HomepageVisitSchema = new Schema(
+  {
+    visitCount: {
+      type: Number,
+      default: 0
+    }
+  },
```
evanugarte: can we add one more field, its called `type`, and it can be an enum like `['HOME_PAGE']` for now

we can then add more stuff to this schema later if we want to track visits to other pages

api/main_endpoints/routes/Homepage.js
```diff
@@ -0,0 +1,34 @@
+const express = require ('express');
+const router = express.Router();
+const visitCounter = require('../models/HomepageVisit');
+const { FORBIDDEN, BAD_REQUEST, OK } = require('../../util/constants').STATUS_CODES;
+const allowedOrigin = 'https://sce.sjsu.edu';
```
evanugarte: what about if its in dev mode? could we only check if we are in prod?

like `if (process.env.NODE_ENV === 'production')`

also lets name this variable `SCE_SJSU_EDU`

src/APIFunctions/HomepageVisit.js
```diff
@@ -0,0 +1,40 @@
...
+    status.responseData = err;
+    status.error = true;
+  }
+  return status;
+}
+
+export async function getVisitCount() {
+  let status = new ApiResponse();
+  try {
+    const res = await fetch (BASE_API_URL + 'api/Homepage/count');
```
evanugarte: please create and use the url class like below

https://github.com/SCE-Development/Clark/blob/33fc1374a33224bdf9e086fc67b60205cbbba5ce/src/APIFunctions/User.js#L107

src/APIFunctions/HomepageVisit.js
```diff
@@ -0,0 +1,40 @@
+import { ApiResponse } from './ApiResponses';
+import { BASE_API_URL } from '../Enums';
+
+export async function incrementVisitCount() {
+  let status = new ApiResponse();
+  try {
+    await fetch (BASE_API_URL + 'api/Homepage/visit', {
```
evanugarte: see other comment about url class

evanugarte (approved):

charred-70 (approved):
lgtm Waow so cool

PR #1988, opened 2026-01-03
Change ui on led sign page to have a tab, if the user is an officer or admin

Embotic-Wayne:
image

### DIFF
--- a/api/main_endpoints/models/PermissionRequest.js
+++ b/api/main_endpoints/models/PermissionRequest.js
@@ -9,6 +9,11 @@ const PermissionRequestSchema = new Schema(
       ref: 'User',
       required: true,
     },
+    status: {
+      type: String,
+      enum: ['PENDING', 'APPROVED', 'DENIED', 'REVOKED'],
+      default: 'PENDING'
+    },
     type: {
       type: String,
       enum: Object.values(PermissionRequestTypes),
--- a/api/main_endpoints/routes/PermissionRequest.js
+++ b/api/main_endpoints/routes/PermissionRequest.js
@@ -51,7 +51,7 @@ router.get('/', async (req, res) => {
       query.type = type;
     }
 
-    const requests = await PermissionRequest.find(query)
+    const requests = await PermissionRequest.findOne(query)
       .populate('userId', 'firstName lastName email')
       .sort({ createdAt: -1 });
 
--- a/api/main_endpoints/routes/PermissionRequest.js
+++ b/api/main_endpoints/routes/PermissionRequest.js
@@ -51,7 +51,7 @@ router.get('/', async (req, res) => {
       query.type = type;
     }
 
-    const requests = await PermissionRequest.findOne(query)
+    const requests = await PermissionRequest.find(query)
       .populate('userId', 'firstName lastName email')
       .sort({ createdAt: -1 });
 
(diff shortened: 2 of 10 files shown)


Embotic-Wayne:
image

api/main_endpoints/models/PermissionRequest.js
```diff
@@ -14,6 +14,10 @@ const PermissionRequestSchema = new Schema(
       enum: Object.values(PermissionRequestTypes),
       required: true,
     },
+    approved: {
+      type: Boolean,
+      default: false,
+    },
```
evanugarte: lets use an enum here, like PENDING, APPROVED, DENIED

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -17,11 +20,23 @@ router.post('/create', async (req, res) => {
...
+
+    if (existingRequest) {
+      if (existingRequest.status === PermissionRequestStatus.PENDING) {
+        return res.sendStatus(CONFLICT);
+      }
+      existingRequest.status = PermissionRequestStatus.PENDING;
+      existingRequest.createdAt = new Date();
+      await existingRequest.save();
+      return res.sendStatus(OK);
+    }
```
evanugarte: i think if it exists we just return conflict no matter what, what do you think

if its denied, we can set the status to pending

api/main_endpoints/routes/PermissionRequest.js
```diff
@@ -62,6 +81,36 @@ router.get('/', async (req, res) => {
...
 });
 
+router.post('/approve', async (req, res) => {
+  const decoded = await decodeToken(req, membershipState.OFFICER);
+  if (decoded.status !== OK) return res.sendStatus(decoded.status);
+
+  const { type, _id } = req.body;
+  if (!type || !Object.keys(PermissionRequestTypes).includes(type)) {
+    return res.status(BAD_REQUEST).send({ error: 'Invalid type' });
+  }
```
evanugarte: same thing like we did before can we do


```js
`type ${type} is invalid, valid types are ${Object.keys(PermissionRequestTypes)}`
```

src/APIFunctions/PermissionRequest.js
```diff
@@ -40,11 +45,97 @@ export async function createPermissionRequest(type, token) {
...
+    if (res.ok) {
+      // API returns 200 with no body, so we just mark success
+      status.responseData = true;
+    } else if (res.status === 409) {
+      // CONFLICT - duplicate request
+      status.error = true;
+      status.responseData = 'Request already exists';
+    } else {
+      status.error = true;
+    }
```
evanugarte:
```suggestion
    status.error = !!res.ok;
    if (res.status === 409) {
      // CONFLICT - duplicate request
      status.responseData = 'Request already exists';
    }
```

src/APIFunctions/PermissionRequest.js
```diff
@@ -40,11 +45,97 @@ export async function createPermissionRequest(type, token) {
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${token}`,
+      },
+      body: JSON.stringify({ type, _id: id }),
+    });
+
+    if (res.ok) {
+      status.responseData = true;
```
evanugarte: lets not even set responseData

for these create/delete requests we can just rely on if error is true/false for checking if it worked

evanugarte (approved):

weslayer (approved):
W growtopia player

Issue #1967, opened 2025-12-25
[payment project] create an api for handling membership payment confirmation


adarshm11:
the process of verifying who has paid for sce membership is very inefficient. we can fix this by making a lot of the requirements automated using clark. let's help by creating an api that can query which members have applied and paid for membership.

**game plan**

refer to #1965 and #1966 for the other components that will be relevant to this api. our goal is to provide an api that will be called when users try to confirm their membership payment on their profile, and will query the mongodb collection of payments.

here's what the mongodb schema MIGHT look like (may not be exactly the same):
```
userId: the id of the user making the payment request -> will be null until the user is verified
confirmationCode: the randomly generated code that the user has to input to verify
amount: the amount of money they paid as part of their application
status: an enum indicating whether the payment was "Pending", "Completed", or "Rejected"
venmo details: an object dictating the transaction ID, payer name, and other details
```

- [x] set up a file in `/api/main_endpoints/routes`, let's call it `MembershipPayment.js`. then set up an express.js api; you can refer to `OfficeAccessCard.js` in the same directory to see how to set up the file.
- [x] let's create a route called `/verifyMembership`. this will be a POST route that takes in the confirmation code from the frontend. 
- [x] the route needs to do routine checks on permissions and request body. we need to make sure that the user is authenticated and that the POST request body is what we expect. you can refer to the `/delete` route in `OfficeAccessCard.js` to see how to check these things. 
- [x] next, we need to check the database to see if the document we're looking for actually exists. this will require a query of the `MembershipPayments` collection to see if a document exists with `confirmationCode` field being equal to the confirmation code of the user that made the request. we also need to specify in the query that the status of the payment is "pending", so that we can avoid multiple people trying to use the same code.
    - [x] if we cannot find such a document, we should return a 404 not found error to the user
    - [x] database query functions should be defined in `/api/util` directory and imported into your api file
- [x] if we find the document, then we need to update the membership payment document. let's set the `status` field to be "completed", and the `userId` field to be the `_id` of the user who made the request (we can use `decoded.token._id` to get this).
- [x] we also need to update the user's membership state: 
```js
if (amount >= 30) {
  member for a year
} else if (amount >= 20) {
  member for a semester
} else {
  no membership
}
```
[...]

PR #1993, opened 2026-01-05
Emily/20250104 handle membership payment confirmation

codebyemily:
issue #1967

### DIFF
--- a/api/main_endpoints/routes/MembershipPayment.js
+++ b/api/main_endpoints/routes/MembershipPayment.js
@@ -15,74 +15,74 @@ const { decodeToken } = require('../../util/auth');
 const { writeLogToClient } = require('../../util/logging');
 
 router.post('/verifyMembership', async (req, res) => {
-    const decoded = await decodeToken(req, membershipState.PENDING);
-    if (decoded.status !== OK) {
-        return res.sendStatus(decoded.status);
+  const decoded = await decodeToken(req, membershipState.PENDING);
+  if (decoded.status !== OK) {
+    return res.sendStatus(decoded.status);
+  }
+
+  const { confirmationCode } = req.body;
+  const userId = decoded.token._id;
+
+  if (!confirmationCode) {
+    writeLogToClient(req.method, {
+      statusCode: BAD_REQUEST,
+      message: 'please type in your confirmation code',
+    });
+    return res.sendStatus(BAD_REQUEST);
+  }
+
+  try {
+    const paymentDocument = await findPayment(confirmationCode);
+
+    if (!paymentDocument) {
+      writeLogToClient(req.method, {
+        statusCode: NOT_FOUND,
+        message: 'payment not found',
+      });
+      return res.sendStatus(NOT_FOUND);
+    }
+    const paymentId = paymentDocument._id;
+    const amount = paymentDocument.amount;
+    let accessLevel;
+    let membershipValidUntil;
+
+    if (amount >= 30) {
+      accessLevel = membershipState.MEMBER;
+      membershipValidUntil = getMemberExpirationDate(2);
+    } else if (amount >= 20) {
+      accessLevel = membershipState.MEMBER;
+      membershipValidUntil = getMemberExpirationDate(1);
+    } else {
+      await rejectPayment(paymentId);
+      writeLogToClient(req.method, {
+        statusCode: BAD_REQUEST,
+        message: 'not enough money sent to become a member',
+      });
+      return res.sendStatus(BAD_REQUEST);
     }
 
-    const { confirmationCode } = req.body 
-    const userId = decoded.token._id
-
-    if (!confirmationCode) {
-        writeLogToClient(req.method, {
-            statusCode: BAD_REQUEST,
-            message: 'please type in your confirmation code',
-        });
-        return res.sendStatus(BAD_REQUEST);
-    }
-
-    try {
-        const paymentDocument = await findPayment(confirmationCode)
-
-        if (!paymentDocument) {
-            writeLogToClient(req.method, {
-                statusCode: NOT_FOUND,
-                message: 'payment not found',
-            });
-            return res.sendStatus(NOT_FOUND);
+    await verifyPayment(paymentId, userId);
+    await User.updateOne(
+      { _id: userId },
+      {
+        $set: {
+          accessLevel,
+          membershipValidUntil
         }
-        const paymentId = paymentDocument._id
-        const amount = paymentDocument.amount
-        let accessLevel
-        let membershipValidUntil
-        
-        if (amount >= 30) {
-            accessLevel = membershipState.MEMBER
-            membershipValidUntil = getMemberExpirationDate(2)
-        } else if (amount >= 20) {
-            accessLevel = membershipState.MEMBER
-            membershipValidUntil = getMemberExpirationDate(1)
-        } else {
-            await rejectPayment(paymentId)
-            writeLogToClient(req.method, {
-                statusCode: BAD_REQUEST,
-                message: 'not enough money sent to become a member',
-            });
-            return res.sendStatus(BAD_REQUEST);
-        }
+      }
+    );
 
-        await verifyPayment(paymentId, userId)
-        await User.updateOne(
-            { _id: userId },
-            {
-                $set: {
-                    accessLevel,
-                    membershipValidUntil
-                }
-            }
-        )
+    writeLogToClient(req.method, {
+      statusCode: OK,
+      message: 'successfully verified user',
+    });
+    return res.sendStatus(OK);
 
-        writeLogToClient(req.method, {
-            statusCode: OK,
-            message: 'successfully verified user',
-        });
-        return res.sendStatus(OK);
-
-    } catch (error){
-        writeLogToClient(req.method, {
-            statusCode: SERVER_ERROR,
-            message: 'internal server error',
-        });
-        return res.sendStatus(SERVER_ERROR);
-    }
-})
+  } catch (error){
+    writeLogToClient(req.method, {
+      statusCode: SERVER_ERROR,
+      message: 'internal server error',
+    });
+    return res.sendStatus(SERVER_ERROR);
+  }
+});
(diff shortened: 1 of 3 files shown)


api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,88 @@
...
+  SERVER_ERROR,
+  NOT_FOUND,
+  OK,
+} = require('../../util/constants').STATUS_CODES;
+
+const User = require('../models/User');
+const { getMemberExpirationDate } = require('../util/userHelpers');
+const { findPayment, verifyPayment, rejectPayment } = require('../util/membershipPaymentQueries');
+const { decodeToken } = require('../../util/auth');
+const { writeLogToClient } = require('../../util/logging');
```
adarshm11: `writeLogToClient()` is a function used specifically for Card Reader because we write logs to that page, in this case we will not be using SSE logs so this import and any uses of this function can be safely deleted from this file

api/main_endpoints/util/membershipPaymentQueries.js
```diff
@@ -0,0 +1,31 @@
...
+
+export async function verifyPayment(paymentId, userId) {
+  await MembershipPayment.updateOne(
+    { _id: paymentId },
+    { $set: {
+      userId,
+      status: status.COMPLETED
+    }}
+  );
+}
```
adarshm11: these two functions can be combined into one. we can use the MongoDB `findOneAndUpdate()` function to first look for a document that has the confirmationCode and status = pending, then use the `$set` clause to set userId and status = completed

api/main_endpoints/util/membershipPaymentQueries.js
```diff
@@ -0,0 +1,31 @@
+import MembershipPayment from '../main_endpoints/models/MembershipPayment.js';
+const status = {
+  PENDING: 'pending',
+  COMPLETED: 'completed',
+  REJECTED: 'rejected',
+};
+
+export async function findPayment(confirmationCode) {
```
adarshm11: let's make it so that these functions are imported at the bottom using `module.exports = { function1, function2, ... }` and then the labels can be `function` instead of `export function` -> this pattern should be there in `/api/main_endpoints/util/OfficeAccessCard.js` as a template if you need

api/main_endpoints/util/membershipPaymentQueries.js
```diff
@@ -0,0 +1,31 @@
...
+  );
+}
+
+export async function rejectPayment(paymentId) {
+  await MembershipPayment.updateOne(
+    { _id: paymentId },
+    { $set: { status: status.REJECTED } }
+  );
+}
+
```
adarshm11, line 61: can we have functions in this file return a Promise with a resolve function callback, similar to how they're done in some other util files:
```
function verifyPayment(userId, confirmationCode) {
  return new Promise((resolve) => {
    try {
      ...your logic here...
```

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,88 @@
...
+    await verifyPayment(paymentId, userId);
+    await User.updateOne(
+      { _id: userId },
+      {
+        $set: {
+          accessLevel,
+          membershipValidUntil
+        }
+      }
+    );
```
adarshm11: this logic is exactly right, in the future we can move this to be a utility function in `userHelpers.js` but for now this is fine here

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,65 @@
...
+    const { amount } = paymentDocument;
+    let accessLevel;
+    let membershipValidUntil;
+
+    if (amount >= 30) {
+      accessLevel = membershipState.MEMBER;
+      membershipValidUntil = getMemberExpirationDate(2);
+    } else if (amount >= 20) {
+      accessLevel = membershipState.MEMBER;
+      membershipValidUntil = getMemberExpirationDate(1);
```
adarshm11: we can define `const accessLevel = membershipState.MEMBER` after this if/else block since it's always gonna be member, and the else clause returns early

api/main_endpoints/util/membershipPaymentQueries.js
```diff
@@ -0,0 +1,46 @@
...
+        },
+        {
+          $set: { userId, status: status.COMPLETED },
+        },
+        {
+          new: true,
+          runValidators: true,
+        }
+      ).then(payment => resolve(payment))
+        .catch(() => resolve(null));
```
adarshm11:
```suggestion
      , (error, result) => {
          if (error) // -> resolve null
          if (!result) // -> resolve false since we couldn't find the payment
          return resolve(result) // -> return the document since it's successful
```

adarshm11: in other words let's use the `(error, result) =>` callback instead of .then() and .catch() , this pattern should be there in the OfficeAccessCard file

api/main_endpoints/util/membershipPaymentQueries.js
```diff
@@ -0,0 +1,46 @@
...
+        { _id: paymentId },
+        { $set: { status: status.REJECTED } }
+      ).then(result => resolve(result))
+        .catch(() => resolve(null));
+    } catch (err) {
+      resolve(null);
+    }
+  });
+}
+
```
adarshm11, line 61: let's have this function return true if it's successful and false if it's unsuccessful. also, you can use Mongoose's `findByIdAndUpdate()` if you want to use paymentId to find the document here

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,65 @@
...
+  if (!confirmationCode) {
+    return res.sendStatus(BAD_REQUEST);
+  }
+
+  try {
+    const paymentDocument = await findVerifyPayment(confirmationCode, userId);
+
+    if (!paymentDocument) {
+      return res.sendStatus(NOT_FOUND);
+    }
```
adarshm11: see comment in the util file about `findVerifyPayment()` returning null or false. we need to check:
if paymentDocument === null -> there was a server error (500 error code)
if paymentDocument === false -> we couldn't find the document so return 404 not found
then if we pass those two checks we were successful in finding and updating

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,65 @@
...
+  BAD_REQUEST,
+  SERVER_ERROR,
+  NOT_FOUND,
+  OK,
+} = require('../../util/constants').STATUS_CODES;
+const membershipState = require('../../util/constants').MEMBERSHIP_STATE;
+const User = require('../models/User');
+const { getMemberExpirationDate } = require('../util/userHelpers');
+const { findVerifyPayment, rejectPayment } = require('../util/membershipPaymentQueries');
+const { decodeToken } = require('../../util/auth');
```
adarshm11:
```suggestion
const { decodeToken } = require('../util/token-functions.js');
```

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,75 @@
...
+  const userId = decoded.token._id;
+
+  if (!confirmationCode) {
+    return res.sendStatus(BAD_REQUEST);
+  }
+
+  try {
+    const paymentDocument = await findVerifyPayment(confirmationCode, userId);
+    if (paymentDocument === null){
+      throw new Error('Server error');
```
adarshm11:
```suggestion
      return res.sendStatus(SERVER_ERROR);
```

api/main_endpoints/util/membershipPaymentQueries.js
```diff
@@ -0,0 +1,53 @@
...
+        }
+        if (!result) {
+          return resolve(false);
+        }
+        return resolve(true);
+      }
+    );
+  });
+}
+
```
adarshm11, line 61: for both of these functions, let's make the body of the Promise be within a try/catch, where if we catch an error we return resolve(null)

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    const paymentId = paymentDocument._id;
+    const { amount } = paymentDocument;
+    let membershipValidUntil;
+
+    if (amount < 20){
+      const rejected = await rejectPayment(paymentId);
+      if (rejected === null){
+        throw new Error('Server error');
```
adarshm11:
```suggestion
        return res.sendStatus(SERVER_ERROR);
```

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,75 @@
+const express = require('express');
+const router = express.Router();
```
adarshm11, line 2:
```suggestion
const router = express.Router();
const bodyParser = require('body-parser');
```

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const { confirmationCode } = req.body;
+  const userId = decoded.token._id;
+
+  if (!confirmationCode) {
+    return res.sendStatus(BAD_REQUEST);
+  }
+
+  try {
+    const paymentDocument = await findVerifyPayment(confirmationCode, userId);
```
adarshm11: we can remove the try catch from here, since it's already being try/catched in the util file

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    const paymentId = paymentDocument._id;
+    const { amount } = paymentDocument;
+    let membershipValidUntil;
+
+    if (amount < 20){
+      const rejected = await rejectPayment(paymentId);
+      if (rejected === null){
+        return res.sendStatus(NOT_FOUND);
```
adarshm11:
```suggestion
        return res.sendStatus(SERVER_ERROR);
```

adarshm11 (approved):
excellent work

charred-70 (approved):
lgtm Emily you're my hero

Issue #1950, opened 2025-12-15
add small animations on homepage for mouseover


evanugarte:
when we mouse over buttons and text on the homepage, have them enlarge slightly, for fun

PR #1999, opened 2026-01-14
added mouseover animations to buttons and text on home page

ayue09125:
Solution to issue #1950 
The cursor is not visible during screenshots, but the images in order show: 
<ul>the first header</ul>
<ul>the second header</ul>
<ul>the learn more button</ul>
<ul>the join us button</ul>
all hovered on and slightly enlarged.

These animations work fine with alternative screen sizes as well and do not shift other elements when hovered on.

chrome_HRpbm7AnI5


evanugarte:
we will merge #2002 first then come back to this one

adarshm11 (approved):

charred-70 (approved):
wow so cool

PR #2000, opened 2026-01-15
allow users to verify membership on profile

### LINKED ISSUE #1966: [payment project] modify member profile to prompt for confirmation code for payment
the process of verifying who has paid for sce membership is very inefficient. we can fix this by making a lot of the requirements automated using clark. let's help fix this by modifying the UI of user profiles so that they can verify their membership.

**game plan**
the current UI code is found in `src/Pages/Profile/MemberView/Profile.js`. here's how the UI looks right now:  

Image

- [ ] let's add a button that says `Verify Membership` next to the other buttons. when clicked, this button should:
   - [ ] pop up a confirmation modal on the user's screen (use `src/Components/DecisionModal/ConfirmationModal.js`. you can refer to `src/Pages/CardReader/CardReader.js` to see how confirmation modal is used for deleting office access cards
   - [ ] if the user is already a verified member (membership state >= 1), the confirmation modal should say something like "hey, you're already a member, you can't verify membership!!". so you will need to use `response.accessLevel` to determine what type of membership the user has.
   - [ ] if the user is a nonmember, the confirmation modal should have an input box that prompts the user to enter a confirmation code, as well as a submit button. *the submit button should be disabled if the input box is empty*. and then based on if the confirmation was successful or not, we can display a message to the user; for now, let's just give them a filler message like "you just confirmed your membership!"

feel free to use as much creative freedom as you'd like to create the UI, all that matters is functionality, have fun with your UI/UX 😁

chubby-ducky:
resolves issue #1966

### DIFF
--- a/src/APIFunctions/MembershipPayment.js
+++ b/src/APIFunctions/MembershipPayment.js
@@ -1,5 +1,5 @@
-import { ApiResponse } from "./ApiResponses";
-import { BASE_API_URL } from "../Enums";
+import { ApiResponse } from './ApiResponses';
+import { BASE_API_URL } from '../Enums';
 
 export async function verifyMembershipFromDb(token, confirmationCode) {
   let status = new ApiResponse();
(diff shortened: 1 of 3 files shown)


adarshm11 (approved):
W alina

Issue #1954, opened 2025-12-15
[placeholder] add minecraft font, color and rotation to sce homepage announcements


evanugarte:
<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/ed178d33-1257-40fc-8e96-4af16c3ecdba" />

right now we have

<img width="1131" height="932" alt="Image" src="https://github.com/user-attachments/assets/df8a3028-8060-481a-ab08-a3c68d70d61b" />

similar to #1952 , these changes will likely go together

PR #2003, opened 2026-01-18
daniel/20260117 add minecraft-like text for announcements on the home page

plsdoant:
https://github.com/user-attachments/assets/b4459a6c-000a-43b2-b590-454aaa2813e2


plsdoant:
for issue #1954

Issue #2016, opened 2026-01-29
add rate limiting to membership confirmation codes


adarshm11:
when people pay for membership, they validate their membership on their profile page with a confirmation code that they receive via email. to prevent bad actors from spamming random codes trying to game the system, let's introduce a rate limiting system for these codes.

- [ ] each user is given 5 code attempts. after 5 attempts, they will no longer be able to verify their membership
- [ ] we can do this by storing a hashmap in memory that maps userId -> number of attempts
- [ ] every time a bad attempt is made, increment the number; if/when the user makes a good request and verifies their membership, remove their hashmap entry

PR #2020, opened 2026-02-01
made attemptCount map to track # of attempts

codebyemily:
issue #2016


api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -27,6 +29,12 @@ router.post('/verifyMembership', async (req, res) => {
 
   const { confirmationCode } = req.body;
   const userId = decoded.token._id;
+  const attempts = attemptCount.get(userId) ?? 0;
+
+  if (attempts >= MAX_ATTEMPTS){
+    logger.error('User has made too many verification attempts.');
+    return res.status(429).send('Too many verification attempts.');
```
adarshm11: wanna create a `TOO_MANY_REQUESTS` enum for this 429 error code in /api/util/constants and import it here

adarshm11 (commented):
we should also make it so that after every failed code, it returns to the frontend how many attempts are left, and the frontend modal says "wrong code - X attempts remaining" or something like that

src/Pages/Profile/MemberView/VerifyMembershipModal.js
```diff
@@ -15,6 +15,10 @@ export default function VerifyMembershipModal(props) {
       user.token,
       confirmationCode,
     );
+    if (apiResponse.remainingAttempts != undefined){
+      bannerCallback(`Wrong Code - You have ${apiResponse.remainingAttempts} left. Status Code: ${apiResponse.responseData || 500}`, 'red');
+      return;
+    }
```
adarshm11: if we reach this case we know what kind of error it is, we don't need to give them a status code

src/APIFunctions/MembershipPayment.js
```diff
@@ -13,8 +13,11 @@ export async function verifyMembershipFromDb(token, confirmationCode) {
       },
       body: JSON.stringify({ confirmationCode })
     });
+    const data = await res.json();
     status.responseData = res.status;
     status.error = !res.ok;
+    status.remainingAttempts = data.remainingAttempts;
+    status.message = data.error;
```
adarshm11:
```suggestion
    if (res.status == TOO_MANY_ATTEMPTS) {
      const data = await res.json();
      status.responseData = data;
      status.error = false;
    } else {
      status.error = !res.ok;
    }
```
TOO_MANY_ATTEMPTS might not exist as an enum in the frontend container, if so we can just hardcode an enum within this function for now

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -27,6 +30,15 @@ router.post('/verifyMembership', async (req, res) => {
 
   const { confirmationCode } = req.body;
   const userId = decoded.token._id;
+  const attempts = attemptCount.get(userId) ?? 0;
+
+  if (attempts >= MAX_ATTEMPTS){
+    logger.error('User has made too many verification attempts.');
```
adarshm11:
```suggestion
    logger.error(`User ${userId} has made too many verification attempts.`);
```

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -35,8 +47,12 @@ router.post('/verifyMembership', async (req, res) => {
 
   const paymentDocument = await findVerifyPayment(confirmationCode, userId);
   if (!paymentDocument) {
+    attemptCount.set(userId, attempts + 1);
     logger.error('Error verifying payment for user:', userId);
-    return res.status(NOT_FOUND).send('Error verifying payment.');
+    return res.status(NOT_FOUND).json({
+      error: 'Error verifying payment.',
```
adarshm11: including "error" in the json body isn't necessary since it doesn't give us any insight into what happened and it isn't used on the frontend either

src/Pages/Profile/MemberView/VerifyMembershipModal.js
```diff
@@ -15,6 +15,10 @@ export default function VerifyMembershipModal(props) {
       user.token,
       confirmationCode,
     );
+    if (apiResponse.remainingAttempts != undefined){
```
adarshm11:
```suggestion
    if (apiResponse.remainingAttempts) {
```

src/APIFunctions/MembershipPayment.js
```diff
@@ -13,8 +13,13 @@ export async function verifyMembershipFromDb(token, confirmationCode) {
       },
       body: JSON.stringify({ confirmationCode })
     });
-    status.responseData = res.status;
-    status.error = !res.ok;
+    if (res.status == TOO_MANY_ATTEMPTS) {
```
adarshm11, line 17: is `TOO_MANY_ATTEMPTS` defined in this file, if not you can write at the top of the file with the imports:

`const TOO_MANY_ATTEMPTS = 429;` 

for now

api/main_endpoints/routes/MembershipPayment.js
```diff
@@ -27,6 +30,13 @@ router.post('/verifyMembership', async (req, res) => {
 
   const { confirmationCode } = req.body;
   const userId = decoded.token._id;
+  const attempts = attemptCount.get(userId) ?? 0;
+
+  if (attempts >= MAX_ATTEMPTS){
+      logger.error(`User ${userId} has made too many verification attempts.`);    return res.status(TOO_MANY_REQUESTS).json({
```
adarshm11:
```suggestion
      logger.error(`User ${userId} has made too many verification attempts.`);
      return res.status(TOO_MANY_REQUESTS).json({
```

src/Pages/Profile/MemberView/VerifyMembershipModal.js
codebyemily: i think if remainingAttempts is sent and it turns out there are 0 attempts left it will return false, but I want to check if remainingAttempts was sent over

so if it had 0 attempts left, that would never be displayed

adarshm11: okay in that case it's fine, let's use !== instead of != though

adarshm11 (approved):
address the one comment then get one more approver

charred-70 (approved):
lgtm so tuff vro

PR #2040, opened 2026-03-06
Location directives for sound endpoint

adover06:
3 nginx location directives for sound endpoint.


weslayer:
u spelled monitoring wrong

PR #2041, opened 2026-03-06
Location directives for sound endpoint

adover06:
3 nginx location directives for sound endpoint.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -132,6 +132,7 @@ http {
         location /api {
             proxy_pass http://mainendpoints;
         }
+        
         location / {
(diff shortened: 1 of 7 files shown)


nginx.conf
```diff
@@ -116,10 +116,23 @@ http {
...
+        }
+
+        location ~ ^/sound(/.*)?$ {
+            proxy_pass http://mic-monitoring-server.sce;
+        }
+
         location /api {
             proxy_pass http://mainendpoints;
         }
+
```
evanugarte:
```suggestion
```

PR #2050, opened 2026-03-18
add gated SCEvents components

### LINKED ISSUE #2051: feature gate SCEvents components throughout Clark
**CONTEXT**
- we will add stuff for SCEvents to Clark across various PRs; thus, the SCEvents components will be a WIP
- we don't want members of the public to see incomplete SCEvents stuff
- however... we still want to merge our code. basically, **we need a way to "turn on" or "turn off" certain parts of Clark's code, depending on if the user is a developer of SCEvents or not**

**TODO**
- if we add an SCEvents component, we need to conditionally render it based on if we are an SCEvents developer or not
   - these values are stored in `config.json` --> this lives in the backend AND frontend
- locally, add a new section to your `config.json` (create it if you don't have it already, it's just a JSON object!
   - for the frontend, the file should be in [`/src/config`](https://github.com/SCE-Development/Clark/tree/dev/src/config)
   - for the backend, the file should be in [`/api/config`](https://github.com/SCE-Development/Clark/tree/dev/api/config)

something like
```
...
"SCEvents": {
    "ENABLED": false
  },
...
```

then, in your SCEvents component, import it at the top of the file + use it to conditionally render the SCEvents code

for instance:

```
import config from '../../config/config.json';
...
...
const unauthedRoutes = [
    { title: 'About', route: '/about' },
    { title: 'Projects', route: '/projects' },
    { title: 'Summer Internship', route: '/s/internship', newTab: true },
    ...(config.SCEvents?.ENABLED ? [{ title: 'SCEvents', route: '/events', newTab: true }] : []),
  ];
```

codebyemily:
resolves #2051

SCEvents not enabled
Screenshot 2026-03-20 at 7 34 49 PM

SCEvents enabled: 
Screenshot 2026-03-20 at 7 40 52 PM

Screenshot 2026-03-20 at 7 41 10 PM

Screenshot 2026-03-20 at 9 12 10 PM


src/Components/Navbar/UserNavbar.js
```diff
@@ -14,6 +14,7 @@ export default function UserNavbar(props) {
     { title: 'About', route: '/about' },
     { title: 'Projects', route: '/projects' },
     { title: 'Summer Internship', route: '/s/internship', newTab: true },
+    { title: 'Events', route: '/events', newTab: true },
```
steeevin88: this adds the events page to our navbar, but we still need to hide it through the `config.json`; otherwise, after we merge this PR to Clark, people in the public will see the change (we eventually do want to do this... just after we finish SCEvents!)

you'll need to:
1. import config in this component such that it's accessible (ex. `import config from '../../config/config.json';`
2. conditionally render this route if `SCEvents` is enabled (`...(config.SCEvents?.ENABLED ? [{ title: 'SCEvents', route: '/events', newTab: true }] : []),`

could you also put screenshots of this working in this PR description? you'll need to create a `config.json` locally and enable SCEvents in it

steeevin88 (approved):
awesome, could you include screenshots in the PR description if possible? just so as a reviewer I know it works 😅

src/Pages/Events/Events.js
```diff
@@ -0,0 +1,4 @@
+
+export default function EventsPage() {
```
steeevin88, line 6: is this gated? seemed like was through the screenshots but i don't see it in code (besides the navbar)

if not then we need to gate this with the config value too 😅  otherwise, even if people can't click on the `SCEvents` option in the navbar, they can still just type in the `/events` endpoint and they'll land on this page

i think the simplest way to do this would be to do it here in `Events.js`; import the config again, and also import the default "there's nobody here page". then, something like:

```
if SCEvents.isEnabled {
	return <h1>Events Page</h1>;
} else {
	return <NotFound/>
}
```

- maybe make the above a ternary to just be cleaner
- also not sure if this will pass linting
- could we also update the PR name, maybe just "add gated SCEvents components" or something similar

regardless thanks for the quick turnaround + for the screenshots, helps with reviewing 😄

steeevin88 (changes requested):

src/Pages/Events/Events.js
codebyemily: That makes sense lol, pls check out my latest push!

codebyemily: i redirected it to /notfound because otherwise the navbar will appear in the "no one is here" page, is that alright?

steeevin88: oops yea i meant not found, thanks

adarshm11 (commented):
this is done fine but i would recommend mimicking how cleezy does it [here](https://github.com/SCE-Development/Clark/blob/dev/src/Pages/URLShortener/URLShortener.js). in that case people can run cleezy locally and set `ENABLED: true` in config.json. i assume that this page will eventually have an api side on clark too, and whether or not it's enabled on frontend will be determined something like:

page loads -> useEffect queries api route for loading events -> api reads config.json for enabled value -> returns to frontend either data or enabled = false.

this seems like a good pattern for scevents to follow too, once the api side is built out on Clark

steeevin88 (approved):

steeevin88:
> this is done fine but i would recommend mimicking how cleezy does it [here](https://github.com/SCE-Development/Clark/blob/dev/src/Pages/URLShortener/URLShortener.js). in that case people can run cleezy locally and set `ENABLED: true` in config.json. i assume that this page will eventually have an api side on clark too, and whether or not it's enabled on frontend will be determined something like:
> 
> page loads -> useEffect queries api route for loading events -> api reads config.json for enabled value -> returns to frontend either data or enabled = false.
> 
> this seems like a good pattern for scevents to follow too, once the api side is built out on Clark

makes sense + yeah we'll keep this in mind as we build out more stuff on the Clark side

adarshm11 (approved):

PR #2070, opened 2026-04-02
Added SCEvents ability to create event

maernest04:
PR for issue [2068](https://github.com/SCE-Development/Clark/issues/2068)

Added Clark UI and wiring for SCEvents: officers and admins can open the events page and use a create event flow that talks to the SCEvents HTTP API. Submitted events are stored by SCEvents in its own MongoDB (scevents / events), not Clark’s main API database.

- API client (src/APIFunctions/SCEvents.js): getSCEvents, createSCEvent, configurable base URL via - REACT_APP_SCEVENTS_URL (with a local dev fallback).
- Pages: Events.js (listing + Create CTA for officers/admins), CreateEventPage.js + CreateEventFormQuestionBlock.js (form aligned with SCEvents event/registration payload shape).
- Routing: Registered /events and /events/create in Routes.js with OFFICER_OR_ADMIN access for the create route.
- Nav: Admin sidebar links for SCEvents and Create event when config.SCEvents.ENABLED is true (AdminNavbar.js).
- Auth: Create-event access is officer or admin (consistent with other officer/admin tools), not admin-only.

Screenshots of what this would look like

Screenshot 2026-04-01 at 4 06 18 PM
Screenshot 2026-04-01 at 4 06 21 PM
Screenshot 2026-04-01 at 4 06 25 PM
Screenshot 2026-04-03 at 12 02 25 AM

### DIFF
--- a/src/APIFunctions/SCEvents.js
+++ b/src/APIFunctions/SCEvents.js
@@ -83,29 +83,3 @@ export async function createSCEvent(eventBody) {
   }
   return status;
 }
-
-export async function getEventByID(id) {
-  let status = new ApiResponse();
-
-  try {
-    const url = new URL(`/events/${id}`, SCEVENTS_API_URL);
-    const res = await fetch(url.href, {
-      method: 'GET',
-      headers: {
-        'Content-Type': 'application/json',
-      },
-    });
-
-    if (res.ok) {
-      const result = await res.json();
-      status.responseData = result;
-    } else {
-      status.error = true;
-    }
-  } catch (err) {
-    status.error = true;
-    status.responseData = err;
-  }
-
-  return status;
-}
--- a/src/APIFunctions/SCEvents.js
+++ b/src/APIFunctions/SCEvents.js
@@ -1,35 +1,21 @@
 import { ApiResponse } from './ApiResponses';
 
-export function getSCEventsBaseUrl() {
-  return (
-    (typeof process !== 'undefined' && process.env.REACT_APP_SCEVENTS_URL) ||
-    'http://localhost:8002'
-  );
-}
+const SCEVENTS_BASE =
+  (typeof process !== 'undefined' && process.env.REACT_APP_SCEVENTS_URL) ||
+  'http://localhost:8002';
 
 function eventsUrl(path) {
-  const base = getSCEventsBaseUrl().replace(/\/$/, '');
+  const base = SCEVENTS_BASE.replace(/\/$/, '');
   const p = path.startsWith('/') ? path : `/${path}`;
   return `${base}${p}`;
-}
-
-async function readBodyAsJsonOrText(res) {
-  const text = await res.text();
-  if (!text) {
-    return null;
-  }
-  try {
-    return JSON.parse(text);
-  } catch {
-    return text;
-  }
 }
 
 export async function getAllSCEvents() {
   const status = new ApiResponse();
   try {
     const res = await fetch(eventsUrl('/events/'));
-    status.responseData = await readBodyAsJsonOrText(res);
+    const result = await res.json();
+    status.responseData = result;
     if (!res.ok) {
       status.error = true;
     }
@@ -44,7 +30,8 @@ export async function getEventByID(id) {
   const status = new ApiResponse();
   try {
     const res = await fetch(eventsUrl(`/events/${id}`));
-    status.responseData = await readBodyAsJsonOrText(res);
+    const result = await res.json();
+    status.responseData = result;
     if (!res.ok) {
       status.error = true;
     }
@@ -59,8 +46,7 @@ export async function createSCEvent(eventBody) {
   const status = new ApiResponse();
   status.statusCode = null;
   try {
-    const url = eventsUrl('/events/');
-    const res = await fetch(url, {
+    const res = await fetch(eventsUrl('/events/'), {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
@@ -68,7 +54,7 @@ export async function createSCEvent(eventBody) {
       body: JSON.stringify(eventBody),
     });
     status.statusCode = res.status;
-    const body = await readBodyAsJsonOrText(res);
+    const body = await res.json();
     if (res.ok) {
       status.responseData = body;
     } else {
--- a/src/APIFunctions/SCEvents.js
+++ b/src/APIFunctions/SCEvents.js
@@ -1,55 +1,85 @@
 import { ApiResponse } from './ApiResponses';
 
-const SCEVENTS_API_URL = 'http://localhost:8080';
+export function getSCEventsBaseUrl() {
+  return (
+    (typeof process !== 'undefined' && process.env.REACT_APP_SCEVENTS_URL) ||
+    'http://localhost:8002'
+  );
+}
+
+function eventsUrl(path) {
+  const base = getSCEventsBaseUrl().replace(/\/$/, '');
+  const p = path.startsWith('/') ? path : `/${path}`;
+  return `${base}${p}`;
+}
+
+async function readBodyAsJsonOrText(res) {
+  const text = await res.text();
+  if (!text) {
+    return null;
+  }
+  try {
+    return JSON.parse(text);
+  } catch {
+    return text;
+  }
+}
 
 export async function getAllSCEvents() {
-  let status = new ApiResponse();
-
+  const status = new ApiResponse();
   try {
-    const url = new URL('/events/', SCEVENTS_API_URL);
-    const res = await fetch(url.href, {
-      method: 'GET',
-      headers: {
-        'Content-Type': 'application/json',
-      },
-    });
-
-    if (res.ok) {
-      const result = await res.json();
-      status.responseData = result;
-    } else {
+    const res = await fetch(eventsUrl('/events/'));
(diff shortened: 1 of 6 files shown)


steeevin88:
could we do a few things:

1. link the issue that this resolves in the PR description (let's always do this in the future!)
2. i merged Trista's PR that adds the page contents; can you fix the merge conflicts? we should be keeping her UI + just add your button somewhere (similar to what you did with the blank page)
3. you mentioned the port conflicts + it makes sense; let's make the default port for `SCEvents` something else... i think we should make it `8002` (b/c SCEta is `8001`, and Cleezy is `8000`?). Can you double-check this, and then open a new PR in the `SCEvents` repo to change the default port? This means changing the default `gin` server setup to use port `8002` instead of the default, and changing the port mappings in the `docker-compose` as well. In the end, the server should run in its Docker container's port `8002`, and we will map it to our local port `8002`. lmk if this doesn't make sense

i'll give a closer review after these changes, but had a few questions from a brief look:
- why is there an `SCEvents` tab in the admin page? and a separate `Create Event` one? if the `SCEvents` tab leads to the events page let's just remove it so we don't bloat the admin panel (if we need to navigate to it we can manually change the URL ourselves)
- let's get rid of the base url logic and just assume it's the `localhost` port, i don't think we'll ever be doing the `.env` approach

src/APIFunctions/SCEvents.js
```diff
@@ -15,7 +28,7 @@ export async function getAllSCEvents() {
     });
 
     if (res.ok) {
-      const result = await res.json();
+      const result = await readBodyAsJsonOrText(res);
```
steeevin88: why didn't the old implementation work? just curious to see if we actually need this

steeevin88: if we don't need this then preferably let's remove all the changes here

src/APIFunctions/SCEvents.js
```diff
@@ -27,3 +40,32 @@ export async function getAllSCEvents() {
...
+    if (res.ok) {
+      status.responseData = body;
+    } else {
+      status.error = true;
+      status.responseData = body;
+    }
+  } catch (err) {
+    status.error = true;
+    status.responseData =
+      err && typeof err.message === 'string' ? err.message : String(err);
```
steeevin88: oh i see now lol

i think we can just simplify this, instead of the "grab JSON or text", just assume it's json (with `res.json`)

src/Pages/Events/CreateEventFormQuestionBlock.js
```diff
@@ -0,0 +1,120 @@
+/* eslint-disable camelcase -- SCEvents registration question shape uses snake_case */
+import React from 'react';
+
+function newId() {
```
steeevin88: do we need this helper?

steeevin88 (commented):
big pr so i will probably re-review 😅 

Image

could we also increase the padding here?

src/APIFunctions/SCEvents.js
```diff
@@ -1,9 +1,10 @@
 import { ApiResponse } from './ApiResponses';
 
-const SCEVENTS_API_URL = 'http://localhost:8080';
+/** SCEvents HTTP API (local dev). */
```
steeevin88: can we remove this comment

src/APIFunctions/SCEvents.js
```diff
@@ -1,9 +1,10 @@
...
+export const SCEVENTS_API_URL = 'http://localhost:8002';
 
 export async function getAllSCEvents() {
-  let status = new ApiResponse();
+  const status = new ApiResponse();
```
steeevin88, line 6: remove this change since it's not needed anymore

src/Pages/Events/CreateEventPage.js
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  if (!isOfficerOrAdmin) {
+    return (
+      <div className="m-10">
+        <h1 className="mb-4 text-3xl font-extrabold text-gray-900 dark:text-white">
+          Create event
+        </h1>
+        <p className="text-gray-600 dark:text-gray-300">
+          Only officers and administrators can create events.
```
steeevin88, line 229: can you show what this page looks like?

src/Pages/Events/Events.js
```diff
@@ -116,9 +120,19 @@ export default function EventsPage() {
...
+        <div className="mb-3 flex flex-col gap-4 sm:flex-row sm:items-end sm:justify-between">
+          <h1 className="text-4xl font-bold text-white md:text-5xl">
+            SCEvents
+          </h1>
+          {canCreateEvent && (
+            <Link
+              to="/events/create"
+              className="inline-flex shrink-0 items-center justify-center rounded-xl border border-white/20 bg-white/10 px-4 py-2.5 text-sm font-semibold text-white shadow-sm backdrop-blur transition hover:border-white/30 hover:bg-white/[0.15]"
+            >
+              Create event
```
steeevin88: being super nitpicky lol, could we maybe just have a `+` icon instead of the "Create event" text? i think the UI would just look nicer

steeevin88 (approved):
left some small comments but lgtm

please resolve all the comments once done + ping me again before we merge

src/Pages/Events/Events.js
maernest04: Screenshot 2026-04-03 at 11 27 23 PM

src/Pages/Events/CreateEventPage.js
maernest04: Screenshot 2026-04-03 at 11 24 01 PM

steeevin88: oops i meant like if you're not an officer or admin, what does this UI end up looking like?

steeevin88 (approved):
left a reply on one of the above comments but otherwise lgtm 😄 

please resolve all the other comments so this PR is less cluttered, we can merge after

KhoaNguyen706 (approved):
LGTM

src/APIFunctions/SCEvents.js
```diff
@@ -1,55 +1,71 @@
 import { ApiResponse } from './ApiResponses';
 
-const SCEVENTS_API_URL = 'http://localhost:8080';
+const SCEVENTS_BASE =
```
steeevin88: let's remove this change, the old constant is fine

src/APIFunctions/SCEvents.js
```diff
@@ -1,55 +1,71 @@
...
+    const res = await fetch(eventsUrl('/events/'));
+    const result = await res.json();
+    status.responseData = result;
+    if (!res.ok) {
       status.error = true;
     }
   } catch (err) {
-    status.error = true;
     status.responseData = err;
+    status.error = true;
```
steeevin88, line 16: can we revert this change? unless it led to an error i'd prefer to not cherry pick lines

src/APIFunctions/SCEvents.js
```diff
@@ -1,55 +1,71 @@
 import { ApiResponse } from './ApiResponses';
 
-const SCEVENTS_API_URL = 'http://localhost:8080';
+const SCEVENTS_BASE =
+  (typeof process !== 'undefined' && process.env.REACT_APP_SCEVENTS_URL) ||
+  'http://localhost:8002';
 
-export async function getAllSCEvents() {
-  let status = new ApiResponse();
+function eventsUrl(path) {
```
steeevin88: i think this function is a bit overkill, let's just do the old way if possible? where we use the `SCEVENTS_API_URL` constant and just add the endpoint

src/APIFunctions/SCEvents.js
```diff
@@ -1,55 +1,71 @@
...
-      status.responseData = result;
+      status.responseData = body;
     } else {
       status.error = true;
+      status.responseData = body;
     }
   } catch (err) {
     status.error = true;
-    status.responseData = err;
+    status.responseData =
```
steeevin88: why not just `status.responseData = err;`? just curious, wouldn't they do the same thing?

src/APIFunctions/SCEvents.js
```diff
@@ -1,55 +1,71 @@
...
+  } catch (err) {
+    status.responseData = err;
+    status.error = true;
+  }
+  return status;
+}
 
+export async function createSCEvent(eventBody) {
+  const status = new ApiResponse();
+  status.statusCode = null;
```
steeevin88: this line is unnecessary

src/APIFunctions/SCEvents.js
```diff
@@ -1,55 +1,71 @@
...
-      method: 'GET',
+    const res = await fetch(eventsUrl('/events/'), {
+      method: 'POST',
       headers: {
         'Content-Type': 'application/json',
       },
+      body: JSON.stringify(eventBody),
     });
-
+    status.statusCode = res.status;
```
steeevin88: can't we just simplify it to something like:

```
    const body = await res.json();
    status.statusCode = res.status;
    status.responseData = body;

    if (!res.ok) {
      status.error = true;
    }
```

steeevin88 (changes requested):
re-reviewed

steeevin88 (approved):
nice!

PR #2080, opened 2026-04-12
Add Event Edit Page

KhoaNguyen706:
I changed these:
- add Event Edit page ( just the same format with create page) 
- add the little button edit on the top right corner
- Temporarily grant Admin permission for create and update event
- Authorization will add later
- currently enable the SCE events 
Before:
image
We can edit registeration form too:
image
image
after update event:
image

### DIFF
--- a/src/APIFunctions/SCEvents.js
+++ b/src/APIFunctions/SCEvents.js
@@ -1,8 +1,6 @@
 import { ApiResponse } from './ApiResponses';
 
 const SCEVENTS_API_URL = 'http://localhost:8002';
-
-// ======== ORIGINAL BACKEND API REQUESTS ========
 export async function getAllSCEvents() {
   const status = new ApiResponse();
   try {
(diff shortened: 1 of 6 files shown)


steeevin88:
did you test this? could you include screenshots

KhoaNguyen706:
Yes, I did, It worked for mock data

src/APIFunctions/SCEvents.js
```diff
@@ -2,6 +2,7 @@ import { ApiResponse } from './ApiResponses';
 
 const SCEVENTS_API_URL = 'http://localhost:8002';
 
+// ======== ORIGINAL BACKEND API REQUESTS ========
```
steeevin88: remove this comment

src/Components/context/SceContext.js
```diff
@@ -2,11 +2,26 @@ import React, { createContext, useContext } from 'react';
 
 export const SceContext = createContext({
   user: {},
-  setUser: () => {},
```
steeevin88, line 5: is this change needed? if not please remove so we don't cherry pick

src/Components/context/SceContext.js
```diff
@@ -2,11 +2,26 @@ import React, { createContext, useContext } from 'react';
...
+  setUser: () => { },
   authenticated: false,
-  setAuthenticated: () => {},
```
steeevin88, line 7: same here

src/Components/context/SceContext.js
```diff
@@ -2,11 +2,26 @@ import React, { createContext, useContext } from 'react';
...
+  setAuthenticated: () => { },
 });
 
+// ======== ORIGINAL useSCE LOGIC ========
 export function useSCE() {
   return useContext(SceContext);
 }
+// ==========================================
+
+// ======== MOCKED VERSION FOR TESTING ========
```
steeevin88: again im confused why we have a mock version; was there a reason why we didn't just use Clark locally?

src/Pages/Events/CreateEventPage.js
```diff
@@ -238,8 +238,8 @@ export default function CreateEventPage() {
   return (
     <div className="m-10 max-w-4xl px-4 sm:px-6">
       <div className="mb-8 pt-8 pb-2">
-        <Link to="/events" className="block pb-3 text-sm link link-primary">
```
steeevin88, line 241: hmm... this probably doesn't belong in this PR but it's fine

steeevin88 (changes requested):
left a few comments, in general though im confused why we have mock functions? as opposed to testing with actual Clark

i know it'll take a few extra steps to create the account in local Clark, create event, etc... but i'd prefer to be extra and ensure things work


also, i don't think any of this stuff is gated. remember that anything SCEvents related on Clark shouldn't be made public yet... so we need to ensure your new page is gated. (even though technically it's true that accessing this page is hard b/c you'd need access to the gated Events page... we should still be thorough)

src/APIFunctions/SCEvents.js
```diff
@@ -1,7 +1,6 @@
 import { ApiResponse } from './ApiResponses';
 
 const SCEVENTS_API_URL = 'http://localhost:8002';
-
```
steeevin88, line 4: remove this

steeevin88 (approved):
besides one more comment this LGTM!

just to make sure, this is gated right? if so feel free to merge

steeevin88:
i think the edit event page isn't gated, let's add this:

```
...
const isSCEventsEnabled = config.SCEvents?.ENABLED;
...
  if (!isSCEventsEnabled) {
    return <Redirect to="/notfound" />;
  }
```

similar to this:
https://github.com/SCE-Development/Clark/blob/60fc7d43e2028939ecf37d5236cd85f3857fb439/src/Pages/Events/Events.js#L135-L168


obviously technically the edit page won't be accessible since the main SCEvents page isn't accessible + there will be no events, but we should be cautious regardless

steeevin88 (approved):
nice

c1ndytat (approved):
lgtm

Issue #2093, opened 2026-04-24
Max Attendees fix for SCEvents


maernest04:
Right now, we use -1 as the value that needs to be inputted into max attendees to get a no limit event. However, this isn't the best practice so we should fix it by adding a checkbox instead that says "No limit" to act as our method for having a no limit event.

PR #2094, opened 2026-04-24
Fix event registration checkboxes, add API auth header, and add No Limit checkbox UI

maernest04:
For issue #2093

## **Problem**
1. **Broken Checkboxes:** Registration checkboxes were not rendering because the form builder was stripping out the `answer_options` for checkbox-type questions before sending them to the API.
2. **Missing API Authorization:** Creating events failed because the `createSCEvent` API call wasn't attaching the required JWT `Authorization` header.
3. **Unintuitive Max Attendees UI:** The max attendees input relied on magic numbers (leaving it blank to send `-1` for unlimited attendees), which was confusing and easily broken if users typed invalid inputs.

## **Solution**
1. **Fixed Checkbox Persistence:** Updated the form builder and API payload builder to properly preserve and edit `answer_options` for `checkbox` question types, exactly like we do for multiple choice and dropdowns.
2. **Added Auth Header:** Updated the `createSCEvent` function to accept and attach the JWT token as a Bearer header, and passed the token from `localStorage` into the call.
3. **Added "No Limit" Checkbox:** Replaced the max attendees magic number behavior with a dedicated "No limit" checkbox. If checked, it disables the number input and cleanly sends `-1`. If unchecked, frontend validation ensures the user enters a valid positive integer.

## **Files Changed**
- `src/Pages/Events/CreateEventPage.js`
  - Fixed `toApiRegistrationForm` and `updateQuestionType` to include `checkbox` types.
  - Added frontend validation for positive integer inputs on Max Attendees.
  - Replaced max attendees UI with a clean "No limit" checkbox alongside the number input.
  - Retrieved `jwtToken` from `localStorage` to pass to the API function.
- `src/Pages/Events/CreateEventFormQuestionBlock.js`
  - Added `checkbox` to the condition that renders the editable answer options list so admins can define checkbox options.
- `src/APIFunctions/SCEvents.js`
  - Modified `createSCEvent` to accept a `token` argument and insert it into the `Authorization` header.

### DIFF
--- a/src/Pages/Events/CreateEventFormQuestionBlock.js
+++ b/src/Pages/Events/CreateEventFormQuestionBlock.js
@@ -78,7 +78,7 @@ export default function CreateEventFormQuestionBlock({
         </div>
       )}
 
-      {(question.type === 'multiple_choice' || question.type === 'dropdown') && (
+      {(question.type === 'multiple_choice' || question.type === 'dropdown' || question.type === 'checkbox') && (
         <div className="mt-3 space-y-2">
           <span className="text-sm text-gray-500 dark:text-gray-400">Answer options</span>
           {(question.answer_options || []).map((option, optIndex) => (
--- a/src/Pages/Events/CreateEventPage.js
+++ b/src/Pages/Events/CreateEventPage.js
@@ -58,7 +58,7 @@ function toApiRegistrationForm(questions) {
       base.answer_details = { max_chars: q.answer_details.max_chars };
     }
     if (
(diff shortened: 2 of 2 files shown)


steeevin88 (changes requested):
let's rebase this lol, will approve after

steeevin88 (approved):
cool am trusting that you tested this

codebyemily (commented):
lgtm

codebyemily (approved):
lgtm

PR #2105, opened 2026-04-26
Automatically assign door codes to new members via external service

adarshm11:
## Problem
When a user pays for membership, their membership status is updated automatically, but their door code has to be manually assigned. This is a problem because most members pay almost exclusively for the door code so this should be automated.

## Context
I created [DCD](https://github.com/SCE-Development/dcd), a simple microservice to maintain and distribute door codes given to SCE by the department. This engine makes sure that the door codes are distributed evenly; since there are more members than door codes, there are inevitably repeats, but the codes are repeated at an almost identical frequency.

## Implementation
I modified the workflow for membership validation by having Clark query DCD for a new door code, then assigning that door code to the user when updating their membership status. This includes robust error handling for tracking when issues occur with querying DCD, and also made it so that the API call to DCD can be mocked, in case of a local setup where the developer does not have access to the production DCD server.

## Testing
This was tested on a local setup with Clark running in dev mode and DCD running on localhost port 5999. Using a test account with access level `NON_MEMBER`, verifying membership with a dummy payment record successfully updated the user's membership status to `MEMBER` and added a door code to the profile.

## Before Running
Ensure that you set the `DCD_URL` field in `api/config/config.json` to be whatever the address of the DCD service is, so that Clark can correctly query for door codes.

### DIFF
--- a/api/config/config.example.json
+++ b/api/config/config.example.json
@@ -33,7 +33,9 @@
     "ENABLED": false
   },
   "membershipPayment": {
-    "API_KEY": "GO_AWAY_LOL"
+    "API_KEY": "GO_AWAY_LOL",
+    "DCD_ENABLED": false,
+    "DCD_URL": "NOT_SET"
   },
   "secretKey": "super duper secret key",
   "SCEvents": {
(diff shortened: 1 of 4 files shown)


evanugarte (approved):

weslayer (approved):
touching u

PR #2109, opened 2026-04-27
Render event based on user registration status

### LINKED ISSUE #2097: Render event cards differently based on users' registration status
self-explanatory, if someone is registered for an event they should know... or if they got rejected, it should probably say that as well

- not sure what the best way to do this is; you'll need to figure out how to get the attendees of an event, and then check if the current user is an attendee of the event --> this is probably inefficient; maybe the better approach is to make a new backend endpoint, something like "is user registered for this event". but then THAT's super inefficient because you can imagine this endpoint would be called for EVERY single event...
	- you'll need to research efficient ways to determine if a user is a registrant of an event...
	- again maybe it's a new endpoint. maybe it's modifying existing endpoints (ex. getEvents)

- the rejected logic should probably sync with #2091; if they're rejected we should let them join the waitlist if available. up to you if you want to add the rejection as part of this issue or the waitlist one

feel free to make these design decisions yourself

trista-chen-29:
Closes #2097 

### Backend dependency
Relies on SCEvents returning `registration_status` in the events response (JWT-authenticated).

### Changes
- Attach JWT token to SCEvents API requests (`getAllSCEvents`, `getEventByID`)
- Enable user-specific event data (e.g. `registration_status`) 
- Update CalendarView popup:
  - Registered → disabled CTA
  - Pending → disabled CTA
  - Rejected → unavailable state
- Add full-page status views in EventRegistration:
  - Registered
  - Pending
  - Waitlisted (UI support, not full logic)
  - Rejected (UI support, not full logic)
- Refactor status UI into reusable `StatusPanel` component

#### Register
Screenshot 2026-04-27 at 2 27 41 PM
This event was created earlier and has no registration form fields, but users can still register.
Screenshot 2026-04-27 at 2 27 50 PM
Shown immediately after a user submits a registration request.
The duration depends on backend processing time.
Screenshot 2026-04-27 at 2 42 08 PM
Screenshot 2026-04-27 at 2 27 58 PM
Screenshot 2026-04-27 at 3 10 20 PM

#### Registration unavailable
Screenshot 2026-04-27 at 3 38 30 PM

### Notes
- Enables personalized event data via authenticated requests
- Without JWT, registration_status is not returned by backend
- Waitlist and rejection behavior depend on backend logic and related PRs

### DIFF
--- a/src/APIFunctions/SCEvents.js
+++ b/src/APIFunctions/SCEvents.js
@@ -2,10 +2,18 @@ import { ApiResponse } from './ApiResponses';
 import config from '../config/config.json';
 
 const SCEVENTS_API_URL = config.SCEvents?.BASE_URL || '/api/scevents';
-export async function getAllSCEvents() {
+
+export async function getAllSCEvents(token) {
   const status = new ApiResponse();
   try {
-    const res = await fetch(`${SCEVENTS_API_URL}/events/`);
+    const headers = token
+      ? { Authorization: `Bearer ${token}` }
+      : {};
+
+    const res = await fetch(`${SCEVENTS_API_URL}/events/`, {
+      headers,
+    });
+
     const result = await res.json();
     status.responseData = result;
     if (!res.ok) {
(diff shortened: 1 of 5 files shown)


steeevin88 (approved):
nice thanks!

c1ndytat (approved):

Issue #2113, opened 2026-04-29
SCEvents: Allow admin editing


maernest04:
For SCEvents, event creation allowed admins to create and edit their own events but there was no way for other admins to be able to edit events they didn't create. There should be a way for the event's original creator to add and remove admins who can edit their event. 

Check the SCEvents repo issue for more [info](https://github.com/SCE-Development/SCEvents/issues/98)

PR #2114, opened 2026-04-29
Added multi admin support for event creation and editing

maernest04:
# PR for #2113 

## **Problem**
For SCEvents, event creation allowed admins to create and edit their own events but there was no way for other admins to be able to edit events they didn't create. There should be a way for the event's original creator to add and remove admins who can edit their event.

## **Solution**
Added a multi-admin picker UI to both the Create Event and Edit Event pages. Admins can search for other admins by name or email and add/remove them from the event. A new backend endpoint `POST /api/User/admins/validate` resolves stored admin IDs into user objects so the edit page can display names and emails instead of raw IDs.

## **Files Changed**
| File | Change |
|------|--------|
| `src/Pages/Events/CreateEventPage.js` | Multi-admin picker UI, sends admins array in payload |
| `src/Pages/Events/EditEventPage.js` | Multi-admin picker UI, admin ID resolution, self-removal warning |
| `src/APIFunctions/User.js` | `validateEventAdmins` frontend API function |
| `api/main_endpoints/routes/User.js` | `POST /admins/validate` endpoint |
| `test/api/User.js` | Tests for `/admins/validate` |

### Event Admin section of event creation and editing
Screenshot 2026-04-28 at 9 30 35 PM

### Searching for any person to add as an event admin (admin and officers only, there's multiple users with the email test but only the admin shows up

Screenshot 2026-04-28 at 9 30 43 PM

### Warning when attempting to remove every event admin, including themselves
Screenshot 2026-04-28 at 9 31 20 PM

### Warning when attempting to save changes after removing themselves as event admin
Screenshot 2026-04-28 at 9 31 33 PM

### Adding every officer and admin as an event admin
Screenshot 2026-05-01 at 11 52 57 PM

### DIFF
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -128,6 +128,57 @@ router.post('/search', async function(req, res) {
   });
 });
 
+router.post('/admins/validate', async function(req, res) {
+  const decoded = await decodeToken(req, membershipState.NON_MEMBER);
+  if (decoded.status !== OK) {
+    return res.sendStatus(decoded.status);
+  }
+
+  if (!Array.isArray(req.body.ids)) {
+    return res.status(BAD_REQUEST).send({ message: 'ids must be an array.' });
+  }
+
+  const requestedIds = [];
+  const invalidIds = [];
+  const seen = new Set();
+
+  req.body.ids.forEach((id) => {
+    if (typeof id !== 'string' || !id.trim()) {
+      invalidIds.push(id);
+      return;
+    }
+
+    const normalizedId = id.trim();
+    if (seen.has(normalizedId)) {
+      return;
+    }
+    seen.add(normalizedId);
+    requestedIds.push(normalizedId);
+  });
+
+  const objectIdPattern = /^[0-9a-fA-F]{24}$/;
+  const candidateIds = requestedIds.filter((id) => objectIdPattern.test(id));
+  invalidIds.push(...requestedIds.filter((id) => !objectIdPattern.test(id)));
+
+  try {
+    const validAdmins = await User.find({
+      _id: { $in: candidateIds },
+      accessLevel: { $gte: membershipState.OFFICER }
+    }, '_id firstName lastName email accessLevel').lean();
+
+    const validIdSet = new Set(validAdmins.map((user) => user._id.toString()));
+    invalidIds.push(...candidateIds.filter((id) => !validIdSet.has(id)));
+
+    return res.status(OK).send({
+      validAdmins,
+      invalidIds
+    });
+  } catch (err) {
+    logger.error('/admins/validate had an error:', err);
+    return res.sendStatus(BAD_REQUEST);
+  }
+});
+
 // Search for all members
 router.post('/users', async function(req, res) {
   const decoded = await decodeToken(req, membershipState.OFFICER);
@@ -145,6 +196,19 @@ router.post('/users', async function(req, res) {
         }
       }))
     };
+  }
+
+  if (req.body.minRole !== undefined && req.body.minRole !== null) {
+    if (Object.keys(maybeOr).length > 0) {
+      maybeOr = {
+        $and: [
+          maybeOr,
+          { accessLevel: { $gte: req.body.minRole } }
+        ]
+      };
+    } else {
+      maybeOr = { accessLevel: { $gte: req.body.minRole } };
+    }
   }
 
   const sortColumn = req.query.sort || 'joinDate';
--- a/src/APIFunctions/User.js
+++ b/src/APIFunctions/User.js
@@ -13,6 +13,7 @@ export async function getAllUsers({
   page = null,
   sortColumn = null,
   sortOrder = null,
+  minRole = null,
 }) {
   const url = new URL('/api/User/users', BASE_API_URL);
 
@@ -35,12 +36,36 @@ export async function getAllUsers({
       body: JSON.stringify({
         query,
         page,
+        minRole,
       }),
     });
     if (res.ok) {
       const result = await res.json();
       status.responseData = result;
     } else {
+      status.error = true;
+    }
+  } catch(err) {
+    status.error = true;
+  }
+  return status;
+}
+
+export async function validateEventAdmins(token, ids) {
+  let status = new UserApiResponse();
+  const url = new URL('/api/User/admins/validate', BASE_API_URL);
+  try {
+    const res = await fetch(url.href, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${token}`,
+      },
+      body: JSON.stringify({ ids }),
+    });
+    const result = await res.json();
+    status.responseData = result;
+    if (!res.ok) {
       status.error = true;
     }
   } catch(err) {
--- a/src/Pages/Events/CalendarView.js
+++ b/src/Pages/Events/CalendarView.js
@@ -160,12 +160,8 @@ function getBadgeText(event, isAdminView) {
   return '';
 }
 
-function getRegistrationStatus(event) {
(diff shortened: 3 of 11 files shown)


steeevin88 (changes requested):
1. can you add some screenshots to show what the UI looks like?
2. CI is failing, let's fix this
3. from my (albeit brief) understanding, this code only allows us to add Clark admins as event admins, but realistically, we should allow... any person? how much more complicated would this be, i know admins is already a lot of code 😅  thought this would be a simpler change apologies

api/main_endpoints/routes/User.js
```diff
@@ -285,12 +340,16 @@ router.post('/edit', async (req, res) => {
         fieldChanges: JSON.stringify(fieldChanges)
       };
 
-      AuditLog.create({
-        userId: tokenId,
-        action: AuditLogActions.UPDATE_USER,
-        documentId: targetId,
-        details: auditDetails
-      }).catch(logger.error);
```
steeevin88, line 293: let's keep this, this already worked + no need to cherry pick

api/main_endpoints/routes/User.js
```diff
@@ -255,11 +306,15 @@ router.post('/edit', async (req, res) => {
     dataToUpdate.password = hashedPassword;
 
     // Create audit log for password change
-    AuditLog.create({
-      userId: tokenId,
-      action: AuditLogActions.CHANGE_PW,
-      details: { email: existingUser.email, userId: tokenId },
-    }).catch(logger.error);
```
steeevin88, line 262: same here

api/main_endpoints/routes/User.js
```diff
@@ -128,9 +128,60 @@ router.post('/search', async function(req, res) {
...
+  } catch (err) {
+    logger.error('/admins/validate had an error:', err);
+    return res.sendStatus(BAD_REQUEST);
+  }
+});
+
 // Search for all members
 router.post('/users', async function(req, res) {
-  const decoded = await decodeToken(req, membershipState.OFFICER);
+  const decoded = await decodeToken(req, membershipState.NON_MEMBER);
```
steeevin88: ngl i know i said before we should allow non members as event admins but i'd rather not allow non members to search for all members 😓 

let's just keep this as officers then, i think our backend support only officers anyway

src/Pages/Events/EditEventPage.js
```diff
@@ -100,11 +115,25 @@ export default function EditEventPage() {
...
+      const adminResult = await validateEventAdmins(token, adminIds);
+      if (!adminResult.error) {
+        const validAdmins = Array.isArray(adminResult.responseData?.validAdmins)
+          ? adminResult.responseData.validAdmins
+          : [];
+        const validAdminIds = validAdmins.map((admin) => String(admin._id));
+        const unresolvedAdmins = adminIds
+          .filter((adminId) => !validAdminIds.includes(adminId))
+          .map((adminId) => ({ _id: adminId }));
+        setEventAdmins([...validAdmins, ...unresolvedAdmins]);
```
steeevin88: this is useful! but are we rendering it anywhere? just curious, it seems like no

src/Routes.js
```diff
@@ -176,14 +185,6 @@ export const officerOrAdminRoutes = [
...
     allowedIf: allowedIf.OFFICER_OR_ADMIN,
     redirect: '/',
-    inAdminNavbar: true
-  },
-  {
-    Component: EditEventPage,
-    path: '/events/:id/edit',
-    pageName: 'Edit Event',
-    allowedIf: allowedIf.OFFICER_OR_ADMIN,
-    redirect: '/',
```
steeevin88, line 187: could we not keep the old one instead?

steeevin88 (changes requested):

src/APIFunctions/User.js
```diff
@@ -13,6 +13,7 @@ export async function getAllUsers({
   page = null,
   sortColumn = null,
   sortOrder = null,
+  officersOnly = false,
```
steeevin88: this looks strange to me; i think typically this would be a whole new function, ex. `getAllOfficersAndAbove()`. but then this function is a bit complicated, so to make a new similar function would be redudant without extracting some of this function's logic out

i think the best way to do this would be to have a generic role filter passed in, so instead of `officersOnly`, it would be `minRole = null`, and then we would pass in the `minRole` if we want it

im fine if the refactor is in this PR, no need to make a new one

steeevin88 (changes requested):

src/Pages/Events/EditEventPage.js
```diff
@@ -193,6 +270,18 @@ export default function EditEventPage() {
       return;
     }
 
+    if (eventAdminIds.length === 0) {
```
steeevin88: shouldn't we allow this

src/Pages/Events/CreateEventPage.js
```diff
@@ -179,6 +264,10 @@ export default function CreateEventPage() {
       setSubmitError('Could not resolve your user id.');
       return;
     }
+    if (eventAdminIds.length === 0) {
+      setSubmitError('Please select at least one event admin.');
+      return;
+    }
```
steeevin88: hm... should we have this? if we have no admins then ALL of Clark's admins are admins of this event...

does SCEvents handle this? if an events admins change to no admins then does SCEvents automatically know to allow all of Clark's admins to be an event admin?


that said this is a bit complicated + maybe not a useful feature so if you wanna ignore this then lmk

src/Routes.js
```diff
@@ -92,6 +92,8 @@ export const memberRoutes = [
     allowedIf: allowedIf.MEMBER,
     redirect: '/',
   },
+
+
```
steeevin88, line 97: rmeove cherry picked whitespace

src/Routes.js
```diff
@@ -176,15 +178,15 @@ export const officerOrAdminRoutes = [
     pageName: 'Create Event',
     allowedIf: allowedIf.OFFICER_OR_ADMIN,
     redirect: '/',
-    inAdminNavbar: true
+    inAdminNavbar: false
   },
+
```
steeevin88, line 184: same here

steeevin88 (changes requested):
some more small stuff

src/Pages/Events/EditEventPage.js
maernest04: We shouldn't cause that means we'd have 0 admins for an event, if we 0 admins on an event then no person can edit the event. Then we're stuck in a situation of "no one can edit the event and no one can add people as event admins"

src/Pages/Events/CreateEventPage.js
maernest04: We should have this because SCEvents doesn’t treat an empty admins list as “every Clark admin is an event admin”—IsAdmin is still false for everyone, and CanEdit only allows site-role admin to modify the event when admins is empty, which is a fallback. On create, the creator is added to the event with the function ensureCreatorIsEventAdmin. Again, there's an explicit safeguard of "ensure at least one admin" to an event.

src/Pages/Events/EditEventPage.js
steeevin88: but i thought 0 admins = all admins on Clark are event admins? can we double check this

src/Pages/Events/CalendarView.js
```diff
@@ -160,8 +160,12 @@ function getBadgeText(event, isAdminView) {
   return '';
 }
 
-function getRegistrationCta(event) {
-  const registrationStatus = event?.registration_status || 'none';
+function getRegistrationStatus(event) {
```
steeevin88: we don't need this helper, let's just write it inline

steeevin88 (approved):
small nit but loks fine, thanks!

c1ndytat (approved):
wordle wordle wordle

PR #2119, opened 2026-05-01
Pass explicit date range when fetching events

steeevin88:
**CONTEXT**
- Events fetching was broken... the frontend wasn't asking for the right range, which led to events not appearing despite being in the same month

**CHANGES**
- explicit event date range instead of calling /events/ with no query params
- events that are fetched are in the visible month range

**TESTING**
before:
image

after:
image


adarshm11 (approved):

KhoaNguyen706 (approved):
LTTM

Issue #2120, opened 2026-05-01
Support event deletion


steeevin88:
basically, we should delete events. old past events should not get deleted; they should automatically get marked as closed (not sure if this logic is there yet)

however, we should probably have some safeguards to ensure deletion isn't super rampant, for instance:
- confirmation before deletions
- event must be closed OR a draft
- maybe more if you can think of any...

PR #2131, opened 2026-05-02
Added event deletion functionality

maernest04:
# PR for #2120 

## **Problem:** 
Admins were unable to delete events from the UI. To delete events from SCEvents we had to manually send commands to remove them from Redis and MongoDB.

## **Solution:** 
Implement a secure deletion flow that includes user confirmation and status validation, ensuring only non-published events can be removed through the interface.

## **Files Changed:**
- `src/APIFunctions/SCEvents.js`
- `src/Pages/Events/EditEventPage.js`
- `src/Pages/Events/EventEditorForm.js`

## Screenshot of the danger zone delete event button
Screenshot 2026-05-02 at 12 27 46 AM

### DIFF
--- a/src/Pages/Events/EditEventPage.js
+++ b/src/Pages/Events/EditEventPage.js
@@ -2,7 +2,7 @@
 import { useMemo, useState, useEffect, useRef } from 'react';
 import { Link, useHistory, useParams } from 'react-router-dom';
 import { useSCE } from '../../Components/context/SceContext';
-import { getEventByID, updateSCEvent } from '../../APIFunctions/SCEvents';
+import { deleteSCEvent, getEventByID, updateSCEvent } from '../../APIFunctions/SCEvents';
 import { getAllUsers, validateEventAdmins } from '../../APIFunctions/User';
 import { membershipState } from '../../Enums';
 import { toApiRegistrationForm, useEventQuestions } from './useEventQuestions';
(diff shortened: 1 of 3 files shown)


src/Pages/Events/EditEventPage.js
```diff
@@ -236,6 +238,23 @@ export default function EditEventPage() {
     history.push('/events');
   }
 
+  async function handleConfirmDelete() {
+    setDeleteError('');
+    setDeleteSubmitting(true);
+    const token = window.localStorage.getItem('jwtToken');
```
adarshm11, line 244: i might be wrong but i believe there is a hook called `useSCE` which lets you extract the token without needing to directly access local storage. see [here](https://github.com/SCE-Development/Clark/blob/dev/src/Pages/CardReader/CardReader.js#L18). if the token you're accessing is different from the clark jwt token then forget i said any of this

c1ndytat (approved):

src/Pages/Events/EditEventPage.js
steeevin88: this is true lol + we should probably update all refs in SCEvents in Clark

can you open a new PR to do this @maernest04 ? we can just merge this for now + do the "refactor" in a new focused PR

steeevin88 (approved):

Issue #2133, opened 2026-05-03
Call useSCE hook in SCEvents


steeevin88:
SCEvents code is poor; we use a lot of references to the user's `token` from localStorage when we should just be retrieving it via the `useSCE()` hook... so let's do this. this pattern is the common pattern for getting auth information for a given user on a webpage

PR #2140, opened 2026-05-03
Switched SCEvents token reading to useSCE hook

maernest04:
# PR for #2133 

## Problem
Before, SCEvent files in Clark were reading a user's token from local storage. However, Clark already has a built in useSCE() hook to get information about a user.

## Solution
Go through each SCEvents file in clark that would access a user's token and switch it over to use useSCE().


c1ndytat (approved):

KhoaNguyen706 (approved):
lgtm

steeevin88 (approved):

PR #2142, opened 2026-05-04
Add SCEvents auth fallback + Update error msg

steeevin88:
second layer of https://github.com/SCE-Development/SCEvents/pull/137/changes just to be thorough

also updates frontend error message to be more prod focused instead of localhost

image


KhoaNguyen706 (approved):
lftm

c1ndytat (approved):

PR #2144, opened 2026-05-04
Add Support Text Link

KhoaNguyen706:
{D82278A9-2C17-4EA3-BD70-96D5532922B3}
The link works rn!!!!

### DIFF
--- a/src/Components/LinkifiedText/LinkifiedText.js
+++ b/src/Components/LinkifiedText/LinkifiedText.js
@@ -1,0 +1,33 @@
+import React from 'react';
+
+const URL_REGEX = /(https?:\/\/[^\s]+)/g;
+
+export default function LinkifiedText({ children }) {
+  if (typeof children !== 'string') {
+    return <>{children}</>;
+  }
+
+  const parts = children.split(URL_REGEX);
+
+  return (
+    <>
+      {parts.map((part, index) => {
+        if (part.match(URL_REGEX)) {
+          return (
+            <a
+              key={index}
+              href={part}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="text-blue-400 hover:text-blue-300 underline underline-offset-2 break-all"
+              onClick={(e) => e.stopPropagation()}
+            >
+              {part}
+            </a>
+          );
+        }
+        return <React.Fragment key={index}>{part}</React.Fragment>;
+      })}
+    </>
+  );
+}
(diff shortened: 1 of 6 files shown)


src/Components/LinkifiedText/LinkifiedText.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+const URL_REGEX = /(https?:\/\/[^\s]+)/g;
```
steeevin88: gemini said this:

Regex Trailing Punctuation: The regex /(https?:\/\/[^\s]+)/g matches any non-whitespace character. If a user types a URL at the end of a sentence like "Check out https://google.com/.", the trailing period will be included in the link, which will likely break the URL.

is this true? if so let's fix

src/Components/LinkifiedText/LinkifiedText.js
```diff
@@ -0,0 +1,33 @@
...
+
+export default function LinkifiedText({ children }) {
+  if (typeof children !== 'string') {
+    return <>{children}</>;
+  }
+
+  const parts = children.split(URL_REGEX);
```
steeevin88: can we add some truncating? if links are too long maybe let's just cut it short and add "..."

gemini said code would look like this:
```
if (part.match(URL_REGEX)) {
  // Truncate text for display if it's too long
  const displayText = part.length > 50 ? `${part.substring(0, 50)}...` : part;

  return (
    <a
      key={index}
      href={part} // Keeps the full URL for the actual link
      target="_blank"
      rel="noopener noreferrer"
      className="text-blue-400 hover:text-blue-300 underline underline-offset-2 break-all"
      onClick={(e) => e.stopPropagation()}
    >
      {displayText}
    </a>
  );
}
```

src/Components/LinkifiedText/LinkifiedText.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
steeevin88: let's remove this import no need

src/Components/LinkifiedText/LinkifiedText.js
```diff
@@ -0,0 +1,33 @@
...
+            >
+              {part}
+            </a>
+          );
+        }
+        return <React.Fragment key={index}>{part}</React.Fragment>;
+      })}
+    </>
+  );
+}
```
steeevin88: on second thought, i think we shouldn't reinvent the wheel + just use a dedicated package for this, ex. https://github.com/gregjacobs/Autolinker.js

can we try that? we should use .parse() instead of .link() i think

steeevin88 (changes requested):
let's try using https://github.com/gregjacobs/Autolinker.js instead

steeevin88 (approved):
can you change the PR title to fit better

miiiyyi (approved):

PR #2159, opened 2026-06-01
Fix Inconsistent Starting Date in Calendar UI

KhoaNguyen706:
closes [#2158 ](https://github.com/SCE-Development/Clark/issues/2158)

On first load of the current month, we find which week row contains today in the grid and only render from that week down so dates like May 31 aren’t clipped off the bottom. The “Today” button does the same; changing month with the arrows or dropdown still shows the full month.

Before: 
image

After fixing:
image

### DIFF
--- a/src/Pages/Events/Calendar/CalendarCell.js
+++ b/src/Pages/Events/Calendar/CalendarCell.js
@@ -10,7 +10,7 @@ export function CalendarCell({ date, isCurrentMonth, isToday, events, onSelectEv
   return (
     <div
       className={[
-        'min-h-[106px] sm:min-h-[124px] border-b border-r border-slate-700/60 p-1.5 transition-colors duration-150',
+        'min-h-[106px] sm:min-h-0 sm:h-full border-b border-r border-slate-700/60 p-1.5 transition-colors duration-150',
         isCurrentMonth ? '' : 'opacity-30',
         isToday ? 'bg-cyan-500/[0.12]' : 'hover:bg-slate-700/35',
       ].join(' ')}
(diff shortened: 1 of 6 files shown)


adarshm11 (commented):
my issue might have been overkill...

i think it makes more sense if we just manually enforce that the ui always shows the entire month, which could be up to 5 weeks. lets change the css to accomplish that, then we can create a chained pr to address any other issues.

update the screenshots for this new fix when you have them , and we can iterate from there

adarshm11:
ACTUALLY i have inspiration, after you finish the stuff i asked for in the prev comment, lets make ui updates so we can view the calendar in day, week, month, or year view (like how google calendar does) BUT do this IN A SEPARATE PR that is branched off of this branch :)

KhoaNguyen706:
> my issue might have been overkill...
> 
> i think it makes more sense if we just manually enforce that the ui always shows the entire month, which could be up to 5 weeks. lets change the css to accomplish that, then we can create a chained pr to address mobile ui.
> 
> update the screenshots for this new fix when you have them , and we can iterate from there



> ACTUALLY i have inspiration, after you finish the stuff i asked for in the prev comment, lets make ui updates so we can view the calendar in day, week, month, or year view (like how google calendar does) BUT do this IN A SEPARATE PR that is branched off of this branch :)

Okay I got it, i have to fix that first and then Mobile Ui and another pr. Will do later. thanks

adarshm11:
actually i think the issue is small enough to where we can address later, i will rewrite issue and close this pr for now

PR #2166, opened 2026-07-02
updated the axios endpoint calls

heatherngyvn:
updated the create and delete axios endpoints as well as the find host link (Clark)


evanugarte (commented):
ty for this,  i think we also have to update
- https://github.com/SCE-Development/Clark/blob/425ae180916b87060e7ff544d9ecfb7837cd9330/api/main_endpoints/util/cleezyHelpers.js#L15 to be /url/find
- https://github.com/SCE-Development/Clark/blob/bd012c11291df93f85a9d98597330936b6a97e32/nginx.conf#L55 to be /url/find
- change the `logger.error` to match the api paths, like instead of /createUrl it mentions /url/create

evanugarte (approved):

adarshm11 (approved):

PR #2174, opened 2026-08-08
Fix invalid `for` attribute on label elements in EmailPreferences.js

marlware:
JSX requires `htmlFor` instead of `for`. At runtime, using the raw HTML attribute `for` name triggers a React DOM property warning.


adarshm11 (approved):
fix merge conflicts

weslayer (approved):

PR #2180, opened 2026-08-09
Fix dark-on-dark text in MembershipForm

marlware:
clark-site-dark-text

Several labels (Semester/Yearly Plan, Membership Application, Select Major/Plan) had no text color set and defaulted to black, thus appearing invisible against the page's always-dark background when the browser isn't in dark mode.

Added `text-gray-100` to match the rest of the card. Light mode stan 'til I die.


adarshm11 (commented):
what does the text look like when the browser is in dark mode? could you provide a screenshot of that too

PR #2181, opened 2026-08-13
Allow bulk creation of SCE events using CSV files

ethanccheng:
Screenshot 2026-08-12 225314

https://github.com/user-attachments/assets/284f83bb-88db-4376-83c5-58402e40304e

Features: Adds import CSV option to create SCE events page. Displays table for data upon file upload. Click create events to publish all events in CSV file. Currently no support for adding registration questions or adding other event admins.

Image is example of a CSV file to create SCE events. Must include headers and each row is a single event. Date and time must be in the same format shown and publish status + visibility must be lowercase. To disable waitlist, put -1 as value. Publish date can be left empty if N/A.

### DIFF
--- a/src/Pages/Events/CreateEventPage.js
+++ b/src/Pages/Events/CreateEventPage.js
@@ -93,7 +93,7 @@ export default function CreateEventPage() {
   const [headersArray, setHeadersArray] = useState([]);
   const [values, setValues] = useState([]);
   const [confirmModal, setConfirmModal] = useState(false);
-  const [modalWarningMessage, setModalWarningMessage] = useState('');
+  const [modalErrorMessage, setModalErrorMessage] = useState('');
   const debounceRef = useRef(null);
   const isFileUpload = useRef(false);
 
@@ -305,6 +305,15 @@ export default function CreateEventPage() {
     'Max Attendees', 'Waitlist', 'Publish Status', 'Visibility', 'Publish Date',
   ];
 
+  function showFileErrorModal(modalMessage) {
+    setFileData([]);
+    setHeadersArray([]);
+    setValues([]);
+    setModalErrorMessage(modalMessage);
+    setConfirmModal(true);
+    isFileUpload.current = false;
+  }
+
   function handleFileUpload(event) {
     const maxFileSize = 10 * 1024 * 1024;
     const file = event.target.files[0];
@@ -312,7 +321,7 @@ export default function CreateEventPage() {
     if (!file) return;
 
     if (file.size > maxFileSize) {
-      setModalWarningMessage(`File size is ${file.size} and exceeds the 10MB limit.`);
+      setModalErrorMessage(`File size is ${file.size} and exceeds the 10MB limit.`);
       setConfirmModal(true);
       event.target.value = '';
       return;
@@ -336,31 +345,22 @@ export default function CreateEventPage() {
           let emptyValueCounter = 0;
 
           if(row.length < required_number_of_columns) {
-            setFileData([]);
-            setHeadersArray([]);
-            setValues([]);
             const actualHeaders = (headersArray[0] || []).map((h) => h.trim());
             const missingHeaders = EXPECTED_HEADERS.filter((h) => !actualHeaders.includes(h));
-            setModalWarningMessage(`Missing required columns: ${missingHeaders.join(', ')}`);
-            setConfirmModal(true);
-            isFileUpload.current = false;
+            showFileErrorModal(`Missing required columns: ${missingHeaders.join(', ')}`);
             return;
           }
           let missingElements = [];
           for (let i = 0; i < row.length; i++) {
+            // index 9 is treated different because it is the only one that can be accepted as empty
             if (i !== 9 && row[i] === '') {
               emptyValueCounter++;
               missingElements.push(i);
             }
           }
           if (emptyValueCounter > 0) {
-            setFileData([]);
-            setHeadersArray([]);
-            setValues([]);
             const missingColumnNames = missingElements.map((i) => headersArray[0][i]);
-            setModalWarningMessage(`Missing required elements: ${missingColumnNames.join(', ')}`);
-            setConfirmModal(true);
-            isFileUpload.current = true;
+            showFileErrorModal(`Missing required columns: ${missingColumnNames.join(', ')}`);
             return;
           }
         }
@@ -376,7 +376,7 @@ export default function CreateEventPage() {
 
   async function handleFileCreateEvent() {
     if (values.length === 0) {
-      setModalWarningMessage('No valid rows to create. Please upload a valid CSV file.');
+      setModalErrorMessage('No valid rows to create. Please upload a valid CSV file.');
       setConfirmModal(true);
       return;
     }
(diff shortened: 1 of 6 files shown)


src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +298,82 @@ export default function CreateEventPage() {
...
   }
 
+  function handleFileUpload(event) {
+    const maxFileSize = 10 * 1024 * 1024;
+    const file = event.target.files[0];
+
+    if (!file) return;
+
+    if (file.size > maxFileSize) {
+      alert('File size exceeds the 10MB limit.');
```
evanugarte: can we use https://github.com/SCE-Development/Clark/blob/f770f0b27cea175ed3a058c66d3b449b6a0a3bbc/src/Components/DecisionModal/ConfirmationModal.js#L3

also in the message, include how many bytes the file is like

> file is size 1234567, this exceeds the 10MB upload limit

src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +298,82 @@ export default function CreateEventPage() {
...
+        setValues(valuesArray);
+      }
+    });
+
+    isFileUpload.current = true;
+  }
+
+  async function handleFileCreateEvent() {
+    for (const row of values) {
+      const waitlistValue = Number(row[6]);
```
evanugarte: could we be 100% sure that the row has at least 7 elements? like a small if statement, 

and small comment for what each column means like

> date,title,description,...

otherwise this happens
```js
a = []
a[4] // undefined
```

src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +300,142 @@ export default function CreateEventPage() {
...
+            const actualHeaders = (headersArray[0] || []).map((h) => h.trim());
+            const missingHeaders = EXPECTED_HEADERS.filter((h) => !actualHeaders.includes(h));
+            setModalWarningMessage(`Missing required columns: ${missingHeaders.join(', ')}`);
+            setConfirmModal(true);
+            isFileUpload.current = false;
+            return;
+          }
+          let missingElements = [];
+          for (let i = 0; i < row.length; i++) {
+            if (i !== 9 && row[i] === '') {
```
evanugarte: could we leave a comment here on why index 9 is treated differently?

src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +300,142 @@ export default function CreateEventPage() {
...
+          if(row.length < required_number_of_columns) {
+            setFileData([]);
+            setHeadersArray([]);
+            setValues([]);
+            const actualHeaders = (headersArray[0] || []).map((h) => h.trim());
+            const missingHeaders = EXPECTED_HEADERS.filter((h) => !actualHeaders.includes(h));
+            setModalWarningMessage(`Missing required columns: ${missingHeaders.join(', ')}`);
+            setConfirmModal(true);
+            isFileUpload.current = false;
+            return;
```
evanugarte: could this be its own function

that clears the data and shows the modal, this helper function could take th emodal warning message a parameter, lmk if this makes sense

its cause we have the same reset logic below for `emptyValueCounter > 0`

evanugarte (commented):
lets also add some screenshots for what the confirm/error modals look like

tysm for this

src/Pages/Events/CreateEventPage.js
ethanccheng: done

src/Pages/Events/CreateEventPage.js
ethanccheng: done

evanugarte (approved):
is there a way we can have an "example csv" that we can download for users to start filling out

like a button called "download example csv", the csv has all the columns

src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +300,175 @@ export default function CreateEventPage() {
     );
   }
 
+  const EXPECTED_HEADERS = [
+    'Event Name', 'Date', 'Time', 'Location', 'Description',
+    'Max Attendees', 'Waitlist', 'Publish Status', 'Visibility', 'Publish Date',
+  ];
```
steeevin88: is it maybe possible to link/sync this to the existing types? in case long term the fields to create an Event change (ex. new field is added), these `EXPECTED_HEADERS` would become outdated right?

if it's not possible no worries, just a consideration

src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +300,175 @@ export default function CreateEventPage() {
...
+    setFileData([]);
+    setHeadersArray([]);
+    setValues([]);
+    setModalErrorMessage(modalMessage);
+    setConfirmModal(true);
+    isFileUpload.current = false;
+  }
+
+  function handleFileUpload(event) {
+    const maxFileSize = 10 * 1024 * 1024;
```
steeevin88: nit: what is this value? obviously not super hard to figure out but maybe leave a comment denoting what this is so people in the future don't need to do the mental math 😓

src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +300,175 @@ export default function CreateEventPage() {
...
+          if(row.length < required_number_of_columns) {
+            const actualHeaders = (headersArray[0] || []).map((h) => h.trim());
+            const missingHeaders = EXPECTED_HEADERS.filter((h) => !actualHeaders.includes(h));
+            showFileErrorModal(`Missing required columns: ${missingHeaders.join(', ')}`);
+            return;
+          }
+          let missingElements = [];
+          for (let i = 0; i < row.length; i++) {
+            // index 9 is treated different because it is the only one that can be accepted as empty
+            if (i !== 9 && row[i] === '') {
```
steeevin88: should we maybe make a row to field mapping/enum? it's not super clear what "index 9"  is + to figure it out requires looking through the potential many columns in the expected CSV format, it'd be nice to know what it is through code (variable naming)

src/Pages/Events/CreateEventPage.js
```diff
@@ -292,6 +300,175 @@ export default function CreateEventPage() {
...
+        [0] - string - name
+        [1] - string - date
+        [2] - string - time
+        [3] - string - location
+        [4] - string - description
+        [5] - string - max attendees
+        [6] - string - waitlist
+        [7] - string - publish status
+        [8] - string - visibility
+        [9] - string - publish date
```
steeevin88: +1 on the enum, this comment is helpful but it might be just more clear if there's an explicity defined enum atop the file? or maybe not lol your call!

steeevin88 (commented):
thanks for making this PR! feel free to push back on any of these comments this was just a quick scan

also one more thing --> would it be helpful to have unit tests / an E2E test suite for this specific feature? just to make sure it doesn't break in the future... not sure if similar tests exists for SCEvents in Clark but it would be cool to have + ensure reliabiltiy of this feature in the future!

src/Pages/Events/CreateEventPage.js
ethanccheng: made enum, is placement in the file alright?

src/Pages/Events/CreateEventPage.js
ethanccheng: done!

src/Pages/Events/CreateEventPage.js
ethanccheng: good suggestion, but unless im missing something i don't think this is doable rn because the current csv file does not take every field from the original form, only a select few

steeevin88 (approved):
thanks for making changes!

PR #2183, opened 2026-08-24
Add bulk user editing for Clark admins

adarshm11:
## Motivation
Every semester, officers change. The default behavior of officer accounts on Clark is that they never expire, so every semester their accounts must be updated by the president. This is a one-by-one editing process. 

## Implementation
This PR accomplishes two tasks:

- [ ] **Adding customizable pagination to the user manager**. This allows the admin to see 10, 20, 50, or all users. 
- [ ] **Bulk user edits**. The admin can now select multiple users and update all of their membership statuses. This action is gated behind the confirmation modal to avoid mistakes in updates. This was implemented via a new `/bulkEdit` endpoint that is more efficient than repeatedly calling `/edit`. 

## Testing
Unit tests were added to verify the modified functionality of user profiles.

### DIFF
--- a/api/main_endpoints/routes/User.js
+++ b/api/main_endpoints/routes/User.js
@@ -29,6 +29,8 @@ const {sendUnsubscribeEmail} = require('../util/emailHelpers');
 const crypto = require('crypto');
 
 const ROWS_PER_PAGE = 20;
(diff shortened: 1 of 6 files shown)


c1ndytat (approved):
lgtm

trista-chen-29 (approved):
Lgtm

PR #2184, opened 2026-08-25
Remove door codes for expired memberships

adarshm11:
## Motivation
When SCE memberships expire, those members are no longer entitled to a door code. Furthermore, door codes are reset by the department every academic year. Clark never addressed these issues, and memberships that expired were still tied to a door code months, and even years, after expiry.

## The Fix
Every time a user logs into Clark, there is a check to see when their membership expires. If it's already expired, the door code is removed from their profile and their status is set back to `NON_MEMBER`. This is the correctly expected behavior of an expired membership, and fixes any potential scenarios where previous members have access to door codes that may or may not work on the room door.


c1ndytat (approved):

plsdoant (approved):
🔥

